PR #9, opened 2015-12-02
fix(deis): remove redundant targetPort

### LINKED ISSUE #3: remove redundant `targetPort` lines
There are several places in the service manifests where `port` and `targetPort` are the same, so `targetPort` could be omitted for simplicity.

bacongobbler:
closes #3


arschles:
oops I forgot to take these out. thanks for fixing this @bacongobbler

PR #11, opened 2015-12-02
ref(deis): rename deis-database to deis-postgres

bacongobbler:
the image name is still to be fixed up as well, which will be renamed in a followup PR

TODO:
- [ ] PR change at deis/workflow
- [ ] change image name at deis/postgres


bacongobbler:
nvm, was confused on the grand scheme of things.

PR #12, opened 2015-12-03
chore(deis): use deis namespace for all manifests

### LINKED ISSUE #2: chart should use "deis" for namespace, not "default"

krancour:
Closes #2.

### DIFF
--- a/deis/manifests/deis-builder-rc.yaml
+++ b/deis/manifests/deis-builder-rc.yaml
@@ -2,6 +2,7 @@ apiVersion: v1
 kind: ReplicationController
 metadata:
   name: deis-builder
+  namespace: deis
   labels:
     heritage: deis
 spec:
--- a/deis/manifests/deis-builder-service.yaml
+++ b/deis/manifests/deis-builder-service.yaml
@@ -2,6 +2,7 @@ apiVersion: v1
 kind: Service
 metadata:
   name: deis-builder
+  namespace: deis
   labels:
     heritage: deis
 spec:
--- a/deis/manifests/deis-database-rc.yaml
+++ b/deis/manifests/deis-database-rc.yaml
@@ -2,6 +2,7 @@ apiVersion: v1
 kind: ReplicationController
 metadata:
   name: deis-database
+  namespace: deis
   labels:
     heritage: deis
 spec:
--- a/deis/manifests/deis-database-service.yaml
+++ b/deis/manifests/deis-database-service.yaml
@@ -2,6 +2,7 @@ apiVersion: v1
 kind: Service
 metadata:
   name: deis-database
+  namespace: deis
   labels:
     heritage: deis
 spec:
--- a/deis/manifests/deis-etcd-discovery-rc.yaml
+++ b/deis/manifests/deis-etcd-discovery-rc.yaml
@@ -2,6 +2,7 @@ kind: ReplicationController
 apiVersion: v1
 metadata:
   name: deis-etcd-discovery
+  namespace: deis
   labels:
     name: deis-etcd-discovery
 spec:
--- a/deis/manifests/deis-etcd-discovery-service.yaml
+++ b/deis/manifests/deis-etcd-discovery-service.yaml
@@ -2,6 +2,7 @@ kind: Service
 apiVersion: v1
 metadata:
   name: deis-etcd-discovery
+  namespace: deis
   labels:
     name: deis-etcd-discovery
     app: deis
--- a/deis/manifests/deis-etcd-discovery-token.yaml
+++ b/deis/manifests/deis-etcd-discovery-token.yaml
@@ -2,5 +2,6 @@ kind: Secret
 apiVersion: v1
 metadata:
   name: deis-etcd-discovery-token
+  namespace: deis
 data:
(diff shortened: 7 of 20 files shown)


bacongobbler:
I think we also need to introduce a deis namespace object, correct? Helm should be able to install them now.

krancour:
Perhaps.  By default namespaces are created just-in-time, along with a default service account and secret token used by that service account... but that can all be disabled.  So if we care about running on as many k8s clusters as possible, then you are probably right.  @technosophos, any thoughts?

technosophos:
One argument for explicitly creating (or maybe intentionally not creating) a namespace has to do with uninstall. Do we want to force-remove the Deis namespace when we uninstall? If so, we probably want the namespace object declared. If not, we probably do not want to declare the namespace.

mboersma:
> Do we want to force-remove the Deis namespace when we uninstall?

I think so--I would generally expect an uninstall operation to leave no trace.

bacongobbler:
+1. Users have complained about us leaving things like deis-builder-data (a data container used in v1 for caching docker builds) after an uninstall, so we should opt for a "delete everything" approach.

krancour:
Ok... so I'll update the PR accordingly, but it leaves me with a question... how do I enforce helm submitting the manifest for the namespace before submitting all the other manifests?

bacongobbler:
That's [handled by helm](https://github.com/helm/helm/blob/master/action/install.go#L20) automatically :)

krancour:
Nice.  I guess my last question before I fix this up is whether we should also be explicitly creating the service account and token for the deis namespace.

mboersma:
> whether we should also be explicitly creating the service account and token for the deis namespace

I think not: that sounds like something site-specific to me. I'm guessing many users will take this chart and tweak it to add the ServiceAccount they have already defined, and if so we would not want to remove that or the token on `helm uninstall`.

krancour:
I think it's not necessarily site-specific.  Depending how the admission controls are set, every new namespace automatically gets a "default" service account and a token used by that service account.  Components like the router count on Kubernetes mounting that token into the container in a well-known location.

If one wants to use a _different_ service account or token, then they have to modify specs to mount that secret themselves.  But keep in mind that even if someone does that, said service account and token would _still_ have to exist within the "deis" namespace.  So I'm not seeing this as a site-specific thing.

I think the quandary for me is more one of, is it safe to assume that admission controls are typically set as I've described above?  Or do we need to explicitly create non-default service accounts and secret tokens within the "deis" namespace?  The Deis components would use these instead of the defaults and it would insulate us from the scenario where admission controls aren't what we'd have wished they were.

I think I've convinced myself.  But is this making sense to anyone else?

krancour:
Ok... I've updated the PR so it explicitly creates a `deis` namespace, and, in case the `NamespaceLifecycle` admission control is not enabled (ordinarily, among other things, this creates the `default` service account in a new namespace), a service account (named `deis`) is also created in this namespace for use by Deis components.  (It turns out I don't have to explicitly create a token for use by the new service account.  That happens automatically.)

Who is using that new, non-default service account at the moment?  No one.  But I will shortly be PR'ing a manifest that adds deis-router to this chart.  _That_ will use it.  I imagine there are other components here that should be using it as well, but I'm not sure which ones-- it would be any ones that talk to the k8s API.  If anyone knows which those are, I can further amend this PR.

mboersma:
I can imagine scenarios where deleting the ServiceAccount on `helm uninstall deis` would be problematic. But more than that, doesn't running in a non-default ServiceAccount require configuration of `kubectl` to operate in that context? For alpha I think making testing more complicated is detrimental, and I'm worried this does that.

I agree that running deis under the "default" ServiceAccount is probably not recommended for an actual enterprise deployment. But I think creating a "deis" ServiceAccount that lives and dies with the product is equivalent to saying "we created this UNIX login you must use to run photoshop." It's too opinionated, too early IMHO.

krancour:
> But more than that, doesn't running in a non-default ServiceAccount require configuration of kubectl to operate in that context?

@mboersma, I think the notion of ServiceAccount is being misunderstood.  The account to which I am referring is not intended for human use.  Its intended for use _by_ services.  The ServiceAccount's associated access token gets mounted into pods so that things running within (e.g. Deis components) can get read only access to the k8s API.

This already happens today.  By default (i.e. using the usual admission controls), every namespace (including the default namespace) gets a default service account.  Without any action on any of our parts, that default service account's token is mounted into every pod in the namespace.

fwiw, this is what enables a line like the following without any additional fuss or ceremony:

```
kubeClient, err := client.NewInCluster()
```

All I am looking to do is _override_ the use of the default service account in the deis namespace and use a specific service account in the deis namespace whose existence we will have ensured.  The reason for this is to account for deployment scenarios wherein the aforementioned admission controls have been altered such that we can't take the existence of a default service account within the deis namespace for granted.

Is this making more sense?  This already has two LGTMs, but I will hold off on merging until you feel more comfortable with this.

mboersma:
I'm expressing @technosophos' concerns here as well, although perhaps we misunderstood. deis-workflow already does the work of reading the (default) mounted token and using it for k8s API auth. Will that mount point change? (Does this imply changes to deis/workflow's code?) Have you tested this change in its entirety to ensure we can still `deis pull` and so on?

mboersma:
Ok, I'm reassured that this doesn't imply any code changes to deis/workflow and it seems to work fine in testing. :+1:

PR #14, opened 2015-12-03
ref(deis): change image destination to deisci/postgres

bacongobbler:
deis/postgres is now deploying the image name as deisci/postgres.
We are still pending discussion if the rest of the chart will need
to be renamed to deis-postgres, but for now this reflects the latest
version of deis/postgres.

See deis/postgres#12


technosophos:
Why does it remove the pull policy, @bacongobbler

bacongobbler:
rebase error. I'll rectify that

bacongobbler:
ok, fixed!

PR #17, opened 2015-12-04
chore(manifests): use imagePullPolicy: Always

krancour:
As long as we continue tagging CI builds as `v2-alpha`, it makes a lot of sense to use `imagePullPolicy: Always` until we're out of alpha and publishing charts that call out a specific released version.

By doing this, we'll always get the _latest_ builds of `v2-alpha` images each time we `helm install deis`, even though older images with the same tag may already exist on our nodes.


gabrtv:
Mutable tags :sadpanda: .. but I agree with the rationale. LGTM.

PR #27, opened 2015-12-09
feat(logger) - adding logger support to the chart

jchauncey:


technosophos:
So here's my plan on this:

Merge this change into the main Deis chart for now.

When we break down the Deis chart into components, we need to discuss whether or not we are going to mark this as required. IIRC, because it uses DaemonSet, it won't be supported on all of the 1.1 clusters (since, IIRC, that feature is turned off by default). If that's not the case, then we should definitely make this part of the full install.

Any objections to that?

Also, if this is ready for review, can you mark it so, @jchauncey

deis/manifests/deis-logger-fluentd-daemon.yaml
```diff
@@ -0,0 +1,40 @@
...
+spec:
+  template:
+    metadata:
+      name: deis-logger-fluentd
+      labels:
+        heritage: deis
+    spec:
+      containers:
+      - name: deis-logger-fluentd
+        image: quay.io/jchauncey/fluentd-kubernetes:syslog
```
technosophos: Before merge, we should put these in the deisci repo. Same with other images.

mboersma: Agreed, the images should come from quay.io/deisci/logger:v2-alpha for the alpha release, and Travis CI or other automation should be pushing them there. Otherwise this looks good to moi.

jchauncey: Yes I want to move these over. Just getting this started.

In deis/manifests/deis-logger-fluentd-daemon.yaml
https://github.com/deis/charts/pull/27#discussion_r47167767:

> +apiVersion: extensions/v1beta1
> +kind: DaemonSet
> +metadata:
> -  name: deis-logger-fluentd
> -  labels:
> -    heritage: deis
>   +spec:
> -  template:
> -    metadata:
> -      name: deis-logger-fluentd
> -      labels:
> -        heritage: deis
> -    spec:
> -      containers:
> -      - name: deis-logger-fluentd
> -        image: quay.io/jchauncey/fluentd-kubernetes:syslog

Before merge, we should put these in the deisci repo. Same with other
images.

—
Reply to this email directly or view it on GitHub
https://github.com/deis/charts/pull/27/files#r47167767.

deis/manifests/deis-logger-fluentd-daemon.yaml
```diff
@@ -0,0 +1,40 @@
+apiVersion: extensions/v1beta1
+kind: DaemonSet
+metadata:
+  name: deis-logger-fluentd
+  labels:
+    heritage: deis
```
krancour: Should also be putting this in the `deis` namespace like the other components.

deis/manifests/deis-logger-rc.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion: v1
+kind: ReplicationController
+metadata:
+  name: deis-logger
+  labels:
+    heritage: deis
```
krancour: Namespace comment applies here as well.

deis/manifests/deis-logger-svc.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: deis-logger
+  labels:
+    heritage: deis
+    app: deis-logger
```
krancour: Namespace comment applies here as well.

jchauncey:
Starting to work on fixing this PR. Updated to use new `deisci` image. We also need to push the fluentd image to `deisci` as well.

jchauncey:
updated to add the namespace label. working on building a fluentd image using alpine. so that should be the last change we need

deis/manifests/deis-logger-fluentd-daemon.yaml
```diff
@@ -0,0 +1,41 @@
+apiVersion: extensions/v1beta1
```
bacongobbler: do we need to install any extra extensions in order for daemonsets to work in k8s v1.1 or is this ready OOTB? Just wondering if there needs to be additions to the installation docs.

jchauncey: Yes you'll need to enable the daemonsets api. I'll make sure the docs get
updated.

jchauncey:
This PR now moves logger into its own chart `deis-logger`

deis-logger/Chart.yaml
```diff
@@ -0,0 +1,12 @@
+name: deis-logger
+home: http://github.com/deis/deis
+version: 0.1.0
+description: Provides a UDP compatiable logger interface for the deis cli
```
mboersma: typo: "compatible"

deis-logger/Chart.yaml
```diff
@@ -0,0 +1,12 @@
...
+maintainers:
+- Jonathan Chauncey <jchauncey@deis.com>
+details: |-
+  NOTE: This chart requires daemonsets to be enabled. Please see https://github.com/kubernetes/kubernetes/blob/master/docs/api.md#enabling-resources-in-the-extensions-group for more information.
+
+  This chart will start a service and replication controller which is needed to fufil requests to the logger component. The deis controller talks to logger over port 8080 and returns the logs to the deis cli. The logger component also listens on port 514 and accepts any syslog compatiable message.
```
mboersma: typo: "fulfill" and "compatible"

deis-logger/Chart.yaml
jchauncey: hah i had those fixed. guess the rebase took the old version

technosophos:
I'm testing this, but not totally sure what the desired outcome should be.

jchauncey:
Deis logs won't work yet. You'll need to curl the URL for the service at
port 8088/fluentd or whatever the container is called

README.md
```diff
@@ -16,6 +16,17 @@ Deis v2 and Helm are changing quickly. Your feedback and participation are more
...
+
+```console
+--runtime-config=extensions/v1beta1=true,extensions/v1beta1/daemonsets=true
+```
+
+you can also do `ENABLE_DAEMONSETS=true` before running the `kube-up.sh` script.
+
+### Installation 
+
 First, add this Chart repo to Helm to install the "deis" chart:
```
krancour, line 30: I feel like this section is making some pretty significant assumptions about _how_ one has gotten from "zero to kubernetes."  Even on our own team, we have at least half a dozen ways of provisioning k8s, with `kube-up` being just one of them.  This is actually a good thing because it's prepping us for a world of k8s ubiquity where not every cluster has been provisioned through the same means... but imo, it means we also need to be sensitive to that in the docs.

Maybe this can at least be qualified with "if you're using `kube-up`... otherwise, consult your Kubernetes distribution's documentation?"

technosophos: This is not true. For alpha, what extensions need to be installed? None unless you are directly using Logger, which is in a separate chart.

krancour:
@jchauncey since this isn't connected to workflow yet and we query the logger's RESTful interface directly to get logs... do you want to consider making logger routable?

If so, consider modifying the service as follows:

```
apiVersion: v1
kind: Service
metadata:
  name: deis-logger
  namespace: deis
  labels:
    routable: "true"
  # ...
  annotations:
    deis.io/routerConfig: |
      {
        "domains": [ "logger" ]
      }
# ...
```

jchauncey:
I have the change to get this hooked up to workflow but was busy yesterday getting the etcd crashloop bug fixed so I could test it too. I will submit that in just a minute

krancour:
Ok... then strike my last suggestion.  Connecting it to workflow is obviously better.

README.md
jchauncey, line 30: yeah its hard to know all the possible ways that someone will provison k8s. I added the only two ways I know that you could enable daemonsets api. However, right now there is no reason to include this section in the main `deis` chart readme. Instead it should be put in the `deis-logger` chart readme.

jchauncey:
So the problem we will have by making this a separate chart is that workflow wont know where to find the logger if its installed first.

technosophos:
@jchauncey is the solution to make that more configurable on workflow, or just document this? (That is, until we get all this chart stuff sorted out IRT daemon sets)

jchauncey:
im not sure what to do. it bother me that logging wouldnt work out of the box with workflow but it also sucks to depend on daemonsets

jchauncey:
I think we have to include logger in the main workflow chart. If we dont then when someone wants to enable logging the have to first install the logger chart and then restart the workflow rc. 

One other possible option is to include the logger service definition in the workflow chart so that the ip and port is set when the workflow rc starts.

bacongobbler:
> I think we have to include logger in the main workflow chart. If we dont then when someone wants to enable logging the have to first install the logger chart and then restart the workflow rc.

I think that's perfectly acceptable for testing on alpha, IMO. Until DaemonSets are a part of the stable API, if we threw this in workflow everyone testing would have bootstrap k8s with the v1beta api extensions. It'll drag down the entire user experience with the rest of the platform.

> One other possible option is to include the logger service definition in the workflow chart so that the ip and port is set when the workflow rc starts.

That could be a happy medium... or people could just install deis-logger before installing workflow. That way the service is installed before workflow boots.

jchauncey:
yeah but in either case they have to enable DaemonSets.

technosophos:
Because of the daemonset thing, I think we will have to accept that the user will need to restart workflow to get logger installed. It's definitely less than optimal, but given that logger's not in the official requirements for alpha, it's acceptable.

This issue will almost certainly be resolved in the Beta timeframe, since k8s 1.2 will have daemonsets enabled without requiring a special flag.

krancour:
> This issue will almost certainly be resolved in the Beta timeframe, since k8s 1.2 will have daemonsets enabled without requiring a special flag.

Does that mean that starting with beta we're targeting only k8s 1.2+?

technosophos:
There is no way you'll get me to commit to that without ever seeing 1.2. ;-)

jchauncey:
Ok so I have heard that logger is both required and not required for alpha. We should get some clarification on that. @slack @gabrtv

gabrtv:
Not required.  This is the current behavior if logger is not running:

```
$ deis logs
Error: 
500 INTERNAL SERVER ERROR
"Error accessing logs for node"
```

I think that is fine for alpha.

jchauncey:
Alright sounds good. Just wanted to make sure =)

technosophos:
That said, we're at the place where logger is working, right? So the chart is totally usable with Alpha.

jchauncey:
I just need to test and submit the PR for the workflow piece and then yeah its good (i hope)

jchauncey:
What components depend on the workflow-service?

bacongobbler:
@jchauncey AFAIK only [builder](https://github.com/deis/builder/blob/5f192e12a7f36dfeb4cf8c18bef51e0f0c7af594/rootfs/etc/confd/templates/receiver#L15) and the router (indirectly) rely on the workflow service.

README.md
```diff
@@ -16,6 +16,19 @@ Deis v2 and Helm are changing quickly. Your feedback and participation are more
...
+```console
+--runtime-config=extensions/v1beta1=true,extensions/v1beta1/daemonsets=true
+```
+
+you can also do `ENABLE_DAEMONSETS=true` before running the `kube-up.sh` script.
+
+### Installation
+
+First, add this Chart repo to Helm to install the "deis" chart:
+>>>>>>> feat(logger) - adding logger support to the chart
```
arschles: looks like this is from a conflict

deis-logger/manifests/deis-logger-fluentd-daemon.yaml
```diff
@@ -0,0 +1,41 @@
...
+            cpu: 100m
+        volumeMounts:
+        - name: varlog
+          mountPath: /var/log
+        - name: varlibdockercontainers
+          mountPath: /var/lib/docker/containers
+          readOnly: true
+        env:
+        - name: "SYSLOG_HOST"
+          value: $(DEIS_LOGGER_PORT_514_UDP_ADDR)
```
arschles, line 32: I love that you can use env vars on startup - :+1:

README.md
jchauncey: fixed.

jchauncey:
can we get this merged? Would make it easier to test https://github.com/deis/workflow/pull/263

jchauncey:
@deis/core-maintainers can we get this merged?

bacongobbler:
this works for me for the time being until 1.2 is out and GKE is using it in prod.

krancour:
I think we better document the _huge_ caveat that this eliminates most zero to k8s options from consideration by users unless they are willing to make an investment in customizing their cluster or are willing to wait for k8s 1.2.

jchauncey:
There is language in the readme of the chart and the in the Chart.yaml that explains that it needs daemonsets to work and links to the documentation.

For coreos based installations all you need to do is change the fleet file that starts the api server and restart it.

Oh it should also be noted that if you install this chart after deis, you will need to bounce the workflow pod.

krancour:
Ah... I just remembered this is a separate chart for now.  So this doesn't impact the installability of the Deis PaaS atop most zero to k8s options.  Strike my complaint.

PR #34, opened 2015-12-14
fix(deis-builder-service.yaml): add target port for the builder RC

arschles:
Fixes https://github.com/deis/builder/issues/27


arschles:
crap, I accidentally snuck in [another fix](https://github.com/arschles/charts-1/commit/83295ed6216a8380854458e7fbec8d5e761aa677). backing it out now and will open a new PR with it

PR #48, opened 2015-12-22
feat(workflow): remove minio-user secret mount from deis-workflow-rc

smothiki:
requires https://github.com/deis/workflow/pull/176


smothiki:
Closing this as we are thinking of generating pre signed URLS

arschles:
see https://github.com/deis/workflow/issues/182

PR #51, opened 2015-12-24
fix(docs): update the client url to the alpha client

iancoffey:
I had to use the alpha client to get things working, seems like this should be updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ $ kubectl get service deis-workflow  # note the IP address for later
 Then open an SSH session to a Kubernetes minion node, install the `deis` client, and register:
 
 ```console
-$ curl -sSL http://deis.io/deis-cli/install.sh | sh
+$ curl -sSL http://deis.io/deis-cli/install-v2-alpha.sh | bash
 $ mkdir $HOME/bin && mv deis $HOME/bin
 $ deis register 10.247.59.157:8000  # or the appropriate CLUSTER_IP, from above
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -29,7 +29,7 @@ $ kubectl get service deis-workflow  # note the IP address for later
 Then open an SSH session to a Kubernetes minion node, install the `deis` client, and register:
 
 ```console
-$ curl -sSL http://deis.io/deis-cli/install.sh | sh
+$ curl -sSL http://deis.io/deis-cli/install-v2-alpha.sh | sh
```
bacongobbler: Let's change the pipe to `bash` so it'll work on Linux. I'd link to the issue on deis/workflow but I'm on mobile :)

helgi:
deis/workflow#153

bacongobbler:
Wingman @helgi to the rescue :+1:

arschles:
looks like a dupe. should this be closed?

bacongobbler:
@arschles where is the duplicate for this?

arschles:
isn't the duplicate change at https://github.com/deis/workflow/pull/153/files#diff-04c6e90faac2675aa89e2176d2eec7d8R45?

bacongobbler:
two separate projects. This changes deis/chart's README where that PR changed deis/workflow's README.

arschles:
sorry, missed that fact. ignore

PR #54, opened 2015-12-29
chore(release): set image pull policy and point at deis registry repos

sgoings:
I'll merge this once the Docker images are in place.

### DIFF
--- a/deis/manifests/deis-builder-rc.yaml
+++ b/deis/manifests/deis-builder-rc.yaml
@@ -16,8 +16,8 @@ spec:
     spec:
       containers:
         - name: deis-builder
-          image: quay.io/deisci/builder:v2-alpha
-          imagePullPolicy: Always
+          image: quay.io/deis/builder:2.0.0-alpha
+          imagePullPolicy: IfNotPresent
           ports:
             - containerPort: 2223
               name: ssh
--- a/_scripts/set-versions
+++ b/_scripts/set-versions
@@ -1,0 +1,30 @@
+#!/usr/bin/env bash
+
+set -eo pipefail -o nounset
+
+if [ ${1:-""} == "-h" ]; then
+  cat <<EOF
+
+  You can set the following variables:
+
+  IMAGE_PREFIX = "" | "quay.io/" | ...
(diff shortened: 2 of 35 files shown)


deis/manifests/deis-builder-rc.yaml
```diff
@@ -16,8 +16,8 @@ spec:
     spec:
       containers:
         - name: deis-builder
-          image: quay.io/deisci/builder:v2-alpha
-          imagePullPolicy: Always
+          image: quay.io/deis/builder:v2-alpha
+          imagePullPolicy: IfNotPresent
```
helgi, line 20: Why change the policy? Should alpha stay up to date with the alpha channel or do we always want full teardowns to get new image?

technosophos, line 20: I think we're viewing v2-alpha as a single version, instead of a channel. @jchauncey had suggested using `canary` as the channel name.

sgoings:
Changed these versions to adhere to semantic versioning rules.

slack:
At the risk of another :bike: :shed:. Post alpha, I'm thinking we create a new set of charts for `deis-dev` which would always point to dev versions of workflow.

That way `helm install deis/deis` will always get the latest "release" and development can continue on  `deis/deis-dev`.

When helm handles versions we can collapse things back (if that ends up making sense).

krancour:
@slack that's a good idea.  Another idea... can we do that _without_ creating a separate repo for it?  I _think_ helm allows us to put more than one chart in a repo.  That being the case, we could add a dev or canary chart right here in this repo.  `helm install deis/deis` would install the released version while `helm install deis/canary` would install the unreleased moving target.

slack:
Yeah, that's what I meant: `deis-dev` and `canary` are analogous in my comment.
- `helm install deis/deis`, repo `deis/charts`, chart `deis`
- `helm install deis/deis-dev`, repo `deis/charts`, chart `deis-dev`

krancour:
Ok... cool.  I just didn't know if a separate repo was on the table or not.  Wanted to avoid that.

PR #55, opened 2015-12-30
fix(deis-builder-rc.yaml): remove DOCKERIMAGE env var

arschles:
requires https://github.com/deis/builder/pull/43

### DIFF
--- a/deis/manifests/deis-builder-rc.yaml
+++ b/deis/manifests/deis-builder-rc.yaml
@@ -26,9 +26,6 @@ spec:
           env:
             - name: "EXTERNAL_PORT"
               value: "2223"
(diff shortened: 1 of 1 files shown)


krancour:
This change has broken installation of the alpha release.  It's a change to the environment variables passed with the container, without any change to the image.

We should probably avoid updating the `deis` manifests right now and instead only update `deis-dev`.  When we have our beta release, `deis-dev` manifests can probably replace `deis` manifests wholesale.

arschles:
ugh, my bad. the `v2-beta` image is changed, but not the official alphas. reverting now

krancour:
@arschles thanks.

PR #56, opened 2015-12-31
fix(readme): add deis namespace to kubectl commands

rimusz:


README.md
```diff
@@ -22,8 +22,8 @@ $ helm repo add deis https://github.com/deis/charts
 $ helm up
 $ helm fetch deis/deis
 $ helm install deis
```
bacongobbler, line 24: We also need to add `--namespace=deis` here.

rimusz, line 24: @bacongobbler this is exactly what this PR fixes

bacongobbler, line 24: yeah but it's missing on this line as well. The official instructions say to use `helm install deis --namespace=deis` but it's not reflected here, just on the `kubectl` commands below. :)

rimusz, line 24: why do we need `helm install deis --namespace=deis` if the chart's manifest files have `namespace: deis` and we have `deis-namespace.yaml` too?

bacongobbler, line 24: it looks like this was changed while I was gone. Disregard

rimusz, line 24: no worries, things get changed a lot :)

PR #41, opened 2015-12-16
fix(*): use standard app labels & selectors, add heritage label where necessary [WIP]

arschles:
This PR changes all manifests to use standard `app` labels and selectors where appropriate. It also adds `heritage` labels where they're missing.

Fixes #40

PR #57, opened 2015-12-31
feat(charts): add Deis-Lite Chart

rimusz:
This is one node Deis-Lite chart with persistent storage and it currently only works on kube-solo App

### DIFF
--- a/deis-lite/manifests/deis-builder-rc.yaml
+++ b/deis-lite/manifests/deis-builder-rc.yaml
@@ -1,0 +1,43 @@
+apiVersion: v1
+kind: ReplicationController
+metadata:
+  name: deis-builder
+  namespace: deis
+  labels:
+    heritage: deis
+spec:
+  replicas: 1
+  selector:
+    name: deis-builder
+  template:
+    metadata:
+      labels:
+        name: deis-builder
+    spec:
+      containers:
+        - name: deis-builder
+          image: quay.io/deis/builder:2.0.0-alpha
+          imagePullPolicy: IfNotPresent
+          ports:
+            - containerPort: 2223
+              name: ssh
+            - containerPort: 3000
+              name: fetcher
+          env:
+            - name: "EXTERNAL_PORT"
+              value: "2223"
+            # This var needs to be passed so that the minio client (https://github.com/minio/mc) will work in Alpine linux
+            - name: "DOCKERIMAGE"
+              value: "1"
+            - name: "POD_NAMESPACE"
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+          volumeMounts:
+            - name: minio-user
+              mountPath: /var/run/secrets/object/store
+              readOnly: true
+      volumes:
+        - name: minio-user
+          secret:
+            secretName: minio-user
--- a/deis-lite/manifests/deis-builder-rc.yaml
+++ b/deis-lite/manifests/deis-builder-rc.yaml
@@ -8,11 +8,11 @@ metadata:
 spec:
   replicas: 1
   selector:
-    name: deis-builder
+    app: deis-builder
   template:
     metadata:
       labels:
-        name: deis-builder
+        app: deis-builder
     spec:
       containers:
         - name: deis-builder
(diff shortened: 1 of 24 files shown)


deis-lite/manifests/deis-builder-rc.yaml
```diff
@@ -0,0 +1,43 @@
...
+kind: ReplicationController
+metadata:
+  name: deis-builder
+  namespace: deis
+  labels:
+    heritage: deis
+spec:
+  replicas: 1
+  selector:
+    name: deis-builder
```
arschles: It'd be good to use `app` for selectors. I'm in the process of moving that way in the other chart as well: https://github.com/deis/charts/pull/41

rimusz: we can merge this one first, then I will make another PR to address `app` for selectors

arschles: I'd like them to just be `app` here

deis-lite/manifests/deis-router-service.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: deis-router
+  namespace: deis
+  labels:
+    heritage: deis
+spec:
+#  type: LoadBalancer
```
sgoings: Did you mean to comment this out?

rimusz: yep, as for now deis-lite works only on kube-solo

deis-lite/manifests/deis-builder-rc.yaml
rimusz: sure, will rebase the PR

sgoings:
I'm struggling a bit with this pull request. I (and many others) have wanted a lite version of Deis for a while now, so I'm glad you're venturing off to tackle this problem... but I'm concerned that this way of implementing deis-lite will cause headaches for us in the long term.

First, for what this PR introduces: It looks like this PR is essentially a copy of all contents of the deis/ chart directory with a few modifications:
- replicas have all been changed to "1" instead of "3"
- volumes have been created to point to host paths of "/data/deis/**some-component**"
- workflow is accessible from port 8000 on the host

If we go down this route we introduce the following problems that particularly make me worried (especially since this PR is going into the deis/charts repo):
- What is our plan to handle divergent changes between the manifests of deis and deis-lite?
- How does deis-lite stay in line with deis releases?
- How do we ensure high quality deis-lite functionality?
- How can we make deis-lite work on **any** single-node kubernetes setup? (instead of just one flavor of kubernetes:  kube-solo)

What if we chose a slightly different way of implementing this... such as using the new functionality introduced in [`helm generate`](https://github.com/helm/helm/issues/309)? Is that possible?

If you couldn't tell by this point... my main worry stems from the fact that this is mostly a duplication of the set of manifests that already exist in the deis dir and all the concerns that arise from duplicated code (double the testing effort, double the release effort, double the maintenance burden, etc. etc.). :smile:

rimusz:
@sgoings this my adoption (first try) of current v2 Deis chart, with a minimal changes to work on kube-solo:
- replicas have all been changed to "1" instead of "3" - as Lite (single host) does not need 3 replicas for etcd, and having just one allows to stop/restart the machine and etcd still will be in tact.
- volumes have been created to point to host paths of "/data/deis/some-component" - as we currently have no other persistent option implement I used host volumes there 
- workflow is accessible from port 8000 on the host - just to make easier to access workflow for now

I already was looking to `helm templates/generate`  which should allow to stay deis-lite stay in line with deis releases, and to have no duplication of the manifests :)

When we have tackled persistent storage for Deis charts, which of course is some challenge for as we need to support GCE, GKE, AWS, Azure, DO and other cloud and bare-metal Kubernetes setups, then Deis-Lite should work there too.

rimusz:
@arschles updated PR with `app` for selectors

slack:
Going to hold off on this one until post beta. Should have a better idea of what we want and need.

Refs deis/deis#4852

rimusz:
Closing this PR, till we have beta version chart

PR #59, opened 2016-01-04
feat(router): add standalone charts for router and router-dev

krancour:
deis/workflow#197 deis/router#72 and #58 should be merged first.


krancour:
fwiw, I'm PR'ing this because I'd like to add installation instructions to deis/router that are useful to folks who are neither hacking on the router nor looking to run the entire Deis PaaS.

PR #67, opened 2016-01-12
feat(workflow): generate secrets for builder and django

helgi:
Requires helm `0.3` for template support

### DIFF
--- a/deis-dev/manifests/deis-builder-rc.yaml
+++ b/deis-dev/manifests/deis-builder-rc.yaml
@@ -37,7 +37,13 @@ spec:
             - name: minio-user
               mountPath: /var/run/secrets/object/store
               readOnly: true
+            - name: builder-key-auth
+              mountPath: /var/run/secrets/api/auth
+              readOnly: true
       volumes:
         - name: minio-user
           secret:
             secretName: minio-user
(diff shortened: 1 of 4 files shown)


deis-dev/manifests/deis-builder-rc.yaml
```diff
@@ -37,7 +37,13 @@ spec:
             - name: minio-user
               mountPath: /var/run/secrets/object/store
               readOnly: true
+            - name: builder-key-auth
+              mountPath: /var/run/secrets/api/auth
```
arschles, line 41: could this be a more descriptive filename like `/var/run/secrets/builder-api/auth`?

helgi, line 41: It can be, `builder-api` feels weird since it is not the API for builder :)

It expands to `/var/run/secrets/api/auth/builder-key` FYI

arschles, line 41: oh, I missed the `mountPath` below. nvm

helgi:
Those not familiar with how the templating works, you will have to run `helm generate deis-dev` to get the secrets before doing `helm install deis-dev`

It's not all fully integrated together in `helm`

PR #74, opened 2016-01-15
feat(router): Add secret bearing dhparam for router

krancour:
This change compliments deis/router#92 (but doesn't need to wait for that to be merged first).

This provides a .pem as a secret for use with nginx's `ssl_dhparam` directive.  This is integral to earning a better grade from [Qualys SSL Labs](https://www.ssllabs.com/ssltest/).  Without this, a well-known nginx default .pem would be used and the score would be lower.

Obvious question: Does this mean everyone using this chart is using the same .pem.  Yes-- but without setting it, they're all using the same (and better known) .pem anyway.  This is a step in the right direction.

People who are truly security obsessed may elect to edit this.  Fwiw, it does _not_ make any sense to use a helm generator to generate one of these .pems because it's _not_ a quick operation.


helgi:
Looks good - We should document really well how people do their own dhparam

krancour:
Ya.  I agree.  It would be through helm chart customizations.  I think a general way to address this is by adding a section to the documentation re: considerations for running in production.  (v1.x had a similar section.)  I'll open a separate ticket for that.

krancour:
Captured in deis/router#93

PR #75, opened 2016-01-16
feat(charts): Add canary charts

krancour:
As soon at @jchauncey's PRs to each component to stop producing beta images on every build and start producing canary images instead gets merged, the `deis-dev` and `router-dev` charts in this repo are going to get really stale really fast because the images they reference won't be getting updated anymore.

This is a proactive countermeasure so that as soon as all those PRs are merged, we have charts available that are tracking the latest builds of all constituent components.

We can have a separate PR later to remove the `deis-dev` and `router-dev` charts once they are rendered useless.

### DIFF
--- a/deis-canary/Chart.yaml
+++ b/deis-canary/Chart.yaml
@@ -1,0 +1,14 @@
+name: deis-dev
+home: https://github.com/deis/workflow
+version: v2-canary
+description: For testing only!
+maintainers:
+- Deis Team <engineering@deis.com>
+details: |-
+  WARNING: this chart is for testing only!  It tracks the latest builds of every
+  constituent component.  Features may not work, the components are not HA, and there are likely
+  to be bugs.
+
+  Deis (pronounced DAY-iss) is an open source PaaS that makes it easy to deploy
+  and manage applications on your own servers. Deis builds on Kubernetes
+  to provide a lightweight, Heroku-inspired workflow.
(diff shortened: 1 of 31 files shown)


jchauncey:
There still needs to be a chart that tracks a stable set of images like
v2-beta. I don't want to be pulling down images that may have issues when
I'm working on another component.

krancour:
There will be... there will still be the deis chart, which, if I am not mistaken, will itself versioned at some point, then there's deis-canary that just tracks HEAD at all times.

jchauncey:
Right but the main deis chart is tracking v2-alpha which I haven't had much
luck with locally. The v2-beta images seem to be more stable. But we still
need a mechanism to install a very specific version of a chart.

krancour:
Right, so hypothetically, when we cut an alpha2 or a beta, the plain "deis" charts are updated to reflect the latest release.  Or, I think there may be plans for helm to support versioning of charts as well, so that eventually you could pick your version when you helm install.  All this PR does is creates a new chart for the "moving targets" that are canary images since that is effectively what "deis-dev" does currently and will cease to do after your other PRs go through.

jchauncey:
Ok so I think we agreed that we aren't doing this right? Basically the builds coming out of travis should be git tagged instead.

bacongobbler:
yes, that's the plan

PR #81, opened 2016-02-02
feat(tests): add e2e test running in gke

sgoings:
remove unnecessary chart-mate:test, download correct version of chart-mate
- also bail if router IP == null
- install jq if necessary

### DIFF
--- a/deis-tests/manifests/deis-tests-pod.yaml
+++ b/deis-tests/manifests/deis-tests-pod.yaml
@@ -10,5 +10,5 @@ spec:
   - name: deis-e2e
     image: quay.io/deisci/deis-e2e:canary
     command: ["/bin/sh", "-c"]
-    args: ["/bin/tests.test || true"]
+    args: ["/bin/tests.test"]
   restartPolicy: Never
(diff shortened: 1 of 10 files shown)


deis-tests/manifests/deis-tests-pod.yaml
```diff
@@ -10,5 +10,5 @@ spec:
   - name: deis-e2e
     image: quay.io/deisci/deis-e2e:canary
     command: ["/bin/sh", "-c"]
-    args: ["/bin/tests.test || true"]
+    args: ["/bin/tests.test"]
```
sgoings, line 13: @mboersma I'm not sure why this was necessary to have... and it prevents us from getting the right exit code from the ginkgo test run...

mboersma, line 13: It may be vestigial from when I tried to implement the chart with a Kubernetes [Job](http://kubernetes.io/v1.1/docs/user-guide/jobs.html), which won't take no for an answer and doesn't respect `restartPolicy: Never`. If it works without `|| true`, then great.

PR #88, opened 2016-02-05
fix(deis-dev/manifests/deis-minio-rc.yaml): add probes to the minio RC

### LINKED ISSUE #86: add readiness and liveness probes to the deis-dev minio RC manifest
After deis/minio#57 is merged and the new container is built

Ref https://github.com/deis/deis/issues/4809

arschles:
and configuration for the server inside the minio container that serves
the probes

Fixes #86 

Can check the box next to `Add health check` (under `Object Storage` > `minio deis/minio`) in https://github.com/deis/deis/issues/4809 when this is done.

### DIFF
--- a/deis-dev/manifests/deis-minio-rc.yaml
+++ b/deis-dev/manifests/deis-minio-rc.yaml
@@ -18,8 +18,24 @@ spec:
         - name: deis-minio
           image: quay.io/deisci/minio:v2-beta
           imagePullPolicy: Always
+          env:
+            - name: HEALTH_SERVER_PORT
+              value: "8082"
           ports:
             - containerPort: 9000
+            - containerPort: 8082
+          livenessProbe:
+            httpGet:
+              path: /healthz
+              port: 8092
+            initialDelaySeconds: 30
+            timeoutSeconds: 1
+          readinessProbe:
+            httpGet:
+              path: /healthz
+              port: 8092
+            initialDelaySeconds: 30
+            timeoutSeconds: 1
           command:
             - boot
           args:
--- a/deis-dev/manifests/deis-minio-rc.yaml
+++ b/deis-dev/manifests/deis-minio-rc.yaml
@@ -27,13 +27,13 @@ spec:
           livenessProbe:
             httpGet:
               path: /healthz
-              port: 8092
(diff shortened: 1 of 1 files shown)


deis-dev/manifests/deis-minio-rc.yaml
```diff
@@ -18,8 +18,24 @@ spec:
...
+              value: "8082"
           ports:
             - containerPort: 9000
+            - containerPort: 8082
+          livenessProbe:
+            httpGet:
+              path: /healthz
+              port: 8092
+            initialDelaySeconds: 30
+            timeoutSeconds: 1
```
mboersma, line 32: Should this default to 5 seconds, or do we intend to make that a pervasive change later.? See deis/workflow#344.

arschles, line 32: @mboersma seems like https://github.com/deis/workflow/issues/344 is specifically targeted at supporting RDS, but in this case the health readiness and liveness probe endpoints _always_ make an s3 request to the local minio server. IMO if that takes longer than 1 second, the service should be considered down. Thoughts?

mboersma, line 32: > if that takes longer than 1 second, the service should be considered down

Makes sense--there probably isn't a "one size fits all" value to use everywhere. :+1:

deis-dev/manifests/deis-minio-rc.yaml
```diff
@@ -18,8 +18,24 @@ spec:
...
+          env:
+            - name: HEALTH_SERVER_PORT
+              value: "8082"
           ports:
             - containerPort: 9000
+            - containerPort: 8082
+          livenessProbe:
+            httpGet:
+              path: /healthz
+              port: 8092
```
bacongobbler: 8082 and 8092?

arschles: ah, thanks. fix incoming

arschles: fixed in https://github.com/arschles/deis-charts/commit/eed9e8d8f7e851786746aee998673332674f747d

PR #92, opened 2016-02-09
ref(charts): Create and use a svc acct per component

### LINKED ISSUE #91: Every Deis component should have its own service account
Ran into this a while back with the router...

Router depends on a service account token mounted at `/var/run/kubernetes/secret/token` in order to access the k8s API.  By default, if you don't explicitly specify a service account in a manifest, the `default` service account for the namespace in question is used.  The rub, however, is that depending on what admission controls are in place any given k8s cluster, `default` service accounts may not be created automatically when new namespaces are created.

To get around this, the `deis` service account was explicitly added to the `deis` namespace.  Router has been using this for quite some time and it works great.  At the time this SA was added, the expectation was that other Deis components might start using it as well.  So far that hasn't happened, although it should.

In talking with @jchauncey about this, he has suggested explicitly creating a service account _per_ component instead (instead of all sharing one).  This will circumvent the condition where the `default` service account was not auto-generated in the `deis` namespace _and_ will have the added benefit that operators can grant more finely-grained permissions to each component.  For example, router simple has no need to write to the API.  Read-only permissions will suffice.  Workflow, by contrast, requires r/w permissions.

fwiw, k8s clusters that are adequately secured aren't something we're really running into often... yet.  But once people start to wrap their heads around what can happen without ACLing in place, more clusters will be doing this, and these changes will help us future-proof ourselves against that.

krancour:
Fixes #91 

cc @jchauncey

### DIFF
--- a/deis-dev/manifests/deis-builder-rc.yaml
+++ b/deis-dev/manifests/deis-builder-rc.yaml
@@ -14,6 +14,7 @@ spec:
       labels:
         app: deis-builder
     spec:
+      serviceAccount: deis-builder
       containers:
         - name: deis-builder
           image: quay.io/deisci/builder:v2-beta
--- a/deis-dev/manifests/deis-builder-service-account.yaml
+++ b/deis-dev/manifests/deis-builder-service-account.yaml
@@ -1,0 +1,7 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: deis-builder
+  namespace: deis
+  labels:
+    heritage: deis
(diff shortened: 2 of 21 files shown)


jchauncey:
you didnt add one for fluentd

krancour:
@jchauncey nice catch.  Missed logger too.  I'm fixing it.

krancour:
@jchauncey fixed.

jchauncey:
i like it =)

PR #96, opened 2016-02-11
ref(builder): remove fetcher port

aledbf:


deis/manifests/deis-builder-rc.yaml
```diff
@@ -21,8 +21,6 @@ spec:
           ports:
             - containerPort: 2223
               name: ssh
-            - containerPort: 3000
-              name: fetcher
           env:
             - name: "EXTERNAL_PORT"
               value: "2223"
```
arschles, line 28: @aledbf let's leave this as-is for now, since it uses the `quay.io/deis/builder:2.0.0-alpha` image, which still has the fetcher in it

aledbf, line 28: @arschles ok

arschles:
@aledbf sorry, I didn't mean that you should close it. I believe that the removal in the `deis-dev` chart is still worthwhile, after deis/builder#171 is merged!

PR #101, opened 2016-02-18
fix(deis-dev): do not use random ascii code for username/password

bacongobbler:
postgres cannot handle random ascii characters as username/password. Reverting to old behaviour for now.


technosophos:
There is a randomAlphaNum template command that you should use instead.

bacongobbler:
@technosophos thanks! I'll follow up with that.

technosophos:
`randAlphaNum`

PR #103, opened 2016-02-18
feat(_scripts): add version bumper for ci

sgoings:
This script takes a few parameters and replaces the docker image
tag the helm chart uses for the component to <version>
(usually a full git hash, but we shorten to use the 7 digit git hash)
within <chart name> with a certain <component name>

### DIFF
--- a/_scripts/chart-version-bumper.sh
+++ b/_scripts/chart-version-bumper.sh
@@ -1,0 +1,35 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
+
+function usage() {
+  cat <<EOF
+$0 <chart name> <component name> <version>
+EOF
+}
+
+if [ $# -ne 3 ]; then
+  usage
+  exit 1
+fi
+
+CHART_NAME="${1}"
+COMPONENT_NAME="${2}"
+VERSION="${3}"
+SHORT_VERSION="${VERSION:0:7}" # short git hashes (what we use in registries) are the 7 characters
+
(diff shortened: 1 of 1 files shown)


_scripts/chart-version-bumper.sh
```diff
@@ -0,0 +1,35 @@
...
+
+# replace the version with VERSION
+file=""
+for file in ${target_file}; do
+  sed -i .bak "s%image: quay.io/deisci/${COMPONENT_NAME}:.*%image: quay.io/deisci/${COMPONENT_NAME}:${SHORT_VERSION}%" "${file}"
+  echo "${CHART_NAME}: set ${COMPONENT_NAME} to ${SHORT_VERSION}"
+done
+
+# remove the backup file
+find . -name "*.bak" | xargs -I {} rm {}
```
arschles, line 32: beware: my bash-fu is weak. possibly stupid question coming

what's the point of writing the backup file in the above `sed` if it's to be removed right away?

sgoings, line 32: When you use sed and the `-i` (modify in place) flag you have to pass a backup file (... with BSD). Gotta love GNU vs BSD differences. :smile: 

https://www.freebsd.org/cgi/man.cgi?query=sed&sektion=&n=1
https://www.gnu.org/software/sed/manual/sed.html

arschles, line 32: ah, got it

PR #106, opened 2016-02-19
feat(deis-logger): use git based tag for logger

sgoings:
Will be automatically updated by Jenkins

### DIFF
--- a/deis-logger/manifests/deis-logger-rc.yaml
+++ b/deis-logger/manifests/deis-logger-rc.yaml
@@ -17,7 +17,7 @@ spec:
       serviceAccount: deis-logger
       containers:
       - name: deis-logger
-        image: quay.io/deis/logger:v2-beta
+        image: quay.io/deisci/logger:git-ba4427c
         imagePullPolicy: Always
         ports:
         - containerPort: 8088


sgoings:
/cc @jchauncey 

Do we want to do this or do we want to split logger into:

deis-logger (reference v2-beta)
deis-logger-dev (reference git sha)

jchauncey:
So I wonder if there should be a deis-logger-dev chart. But some of this hinges on if we are going to fold deis-logger into the deis chart this week.

sgoings:
So we have two options:
- if we require daemonset support in the deis beta release we're going to have to use something other than GKE to run automated tests
- if we don't require daemonset support, then the charts will have to be separated into the deis-logger/deis-logger-dev chart

@slack can you chime in on the DaemonSet decision?

rimusz:
I would stay away from daemonset support till v1.2 is released and GKE supports it

slack:
The release of Deis v2.0.0 requires DaemonSets for logs to function, so we are pretty much defacto requiring DS as a result.

My current thinking (not a dictation!):
1. We should probably merge the charts so day to day development starts to put coverage on logging functionality.
2. We should templatize the charts to disable logging on GKE so e2e tests aren't entirely impacted.
3. We should look at what we will need to do to support DaemonSets on a 1.2 cluster. When the API is promoted we'll possibly need helm to handle v1.1 and v1.2 k8s modes.

rimusz:
+1

sgoings:
I'm waiting to update + merge this until deis e2e tests are green again.

jchauncey:
closing this now that logger is part of `workflow`

PR #109, opened 2016-02-22
fix(logger): Update env vars for finding syslog port and host

jchauncey:
This is in response to changes from @krancour


krancour:
This is much nicer since it doesn't include the port number in the env var names. :+1:

krancour:
This should be merged _before_ #104

krancour:
This should be combined with #107.

jchauncey:
if you want to put this change in your PR I can just delete this one

krancour:
@jchauncey done.  See updates to #107

PR #110, opened 2016-02-22
feat(_scripts): use git sha of e2e test docker image instead of canary

sgoings:


sgoings:
Uses deis/workflow-e2e#56

_scripts/tests/install.sh
```diff
@@ -22,6 +22,7 @@ if [ ! -z $CI ]; then
   echo "INJECTING VERSIONS!"
   echo "-------------------"
   pushd ../../ &> /dev/null
+  bumpver-if-set "deis-tests" "deis-e2e" "${WORKFLOW_E2E_SHA}"
```
vdice, line 25: really tried to hold back... but how about `WORKFLOW_E2E_GIT_SHA` for consistency? :)

sgoings, line 25: Yeah... Was thinking maybe removing the `_GIT_` on the others for brevity's sake via another PR was more what I would do.

PR #117, opened 2016-02-24
ref(deis-dev): change livenessProbe for deis-database

bacongobbler:
to be tested in tandem with deis/postgres#24

### DIFF
--- a/deis-dev/manifests/deis-database-rc.yaml
+++ b/deis-dev/manifests/deis-database-rc.yaml
@@ -29,7 +29,7 @@ spec:
           livenessProbe:
             exec:
               command:
-                - is_master
+                - is_running
             initialDelaySeconds: 30
             timeoutSeconds: 1
           volumeMounts:


mboersma:
@michelleN and I wondered if `readinessProbe` isn't more appropriate here.

helgi:
A `readiness` probe is more appropriate with a good initial delay. For `liveness` we need to decide if we want to potentially restart the pod if pg reports it self as not running

bacongobbler:
Yeah after tossing around the idea a little bit I think it'd be safer to restart the pod if the database reports itself as down. If minio goes down and the database can no longer back up wal logs, it should become unresponsive, which makes the case for a `readinessProbe` but part of me wonders what kind of issues we could hit in prod with postgres that could make it unresponsive otherwise, which makes me want to lean towards the "kill if unresponsive" approach for now. Once we start getting more data/usage from actual clusters then we can tune these settings as we see fit.

helgi:
The readiness probe is more about not letting anything think it is up and running before it is ready :) thus the delay. But I hear ya

mboersma:
I'm fine with merging the `livenessProbe`--the distinction is arguable here and I agree that all these values are likely to be refined post-beta.

PR #119, opened 2016-02-24
fix(fluentd): Remove cpu limits from fluentd

jchauncey:
This should help fix the problem where daemon set pods are dying randomly.


helgi:
Does that mean fluentd could in theory hog up the full CPU if it had an issue?

jchauncey:
Yeah but we don't have limits on anything right now

PR #122, opened 2016-02-25
fix(deis-dev/workflow): increase timeouts for readiness/liveness probes

sgoings:


helgi:
https://github.com/deis/workflow/issues/344

bacongobbler:
closes deis/workflow#418

PR #123, opened 2016-02-25
feat(registry): add object storage templates

### LINKED ISSUE #95: Make 'helm generate' generate the correct credentials for object storage
The person who executes `helm generate` on the `deis/deis` and `deis/deis-dev` charts should be able to specify their object storage location & credentials. It probably makes the most sense to allow them to specify those values in a values file.

technosophos:
This configuration is for registry. However, configuration is generic for other object storage users, as well.

Closes #95

TO TEST:
- `helm up` your Deis charts repo
- `helm fetch deis/deis-dev`
- `cd $(helm_home/workspace/charts/deis-dev)`
- Edit `tpl/objectstorage.toml`. Directions are in the file.
- `helm generate -x manifests deis-dev`

Then you should have installable manifests.


bacongobbler:
should we also remove https://github.com/deis/charts/blob/master/deis-dev/manifests/deis-registry-rc.yaml since that'll get overwritten?

technosophos:
@bacongobbler Good idea. Done.

kmala:
secret should have the namespace "deis"

technosophos:
Thanks, @kmala. Fixed that.

kmala:
In the toml we need to provide the json between ''' ''' instead of " ".On making this change and testing, everything is working fine.

bacongobbler:
This is a nice abstraction point. Is this how we're planning on creating and sharing S3 credentials across components? If so there's a few things I'll have to change in deis/postgres to make it aware of this secret, as well as other components so the user has only one place they need to dump their S3 credentials. If it's not a beta requirement we could just let users set up each component separately (as much as that's a hassle) then clean it up before stable.

deis-dev/tpl/deis-objectstorage-secret.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: objectstorage-keyfile
+  namespace: deis
+  labels:
+    heritage: deis
+  annotations:
+    deis.com/objectstorage: "{{.storage}}"
```
krancour: Elsewhere, we have used deis.io (or subdomains thereof) as a prefix.  I vaguely recall some short discussion about this a couple months back where we had agreed .io was better in this context because it was more open-sourcey.  That aside, now that .io's been used elsewhere, it's more a matter of consistency now.

deis-dev/tpl/objectstorage.toml
```diff
@@ -0,0 +1,46 @@
...
+# on this chart.
+
+# Set the storage backend
+#
+# Valid values are:
+# - filesystem: Store registry on ephemeral disk
+# - s3: Store registry in AWS S3 (configure in S3 section)
+# - azure: Store registry in Azure's object storage
+# - gcs: Store registry in Google Cloud Storage
+# - minio: Store registry on in-cluster Minio server
```
krancour: > However, configuration is generic for other object storage users

Appearance of "registry" in the preceding lines might undermine that intent.  Can this be reworded somehow?

deis-dev/tpl/deis-objectstorage-secret.yaml
technosophos: Okay. Will fix that

deis-dev/tpl/objectstorage.toml
technosophos: I'll change it

technosophos:
@kmala has manually tested, so I'm removing that flag. All other issues have been fixed as well.

krancour:
Love this change.  I always loved stateless platform and this gets us a step closer to the possibility of doing that.  :100:

mboersma:
If I want to use filesystem (default) storage now, how would I do that? @krancour, @kmala, @technosophos: how did you test that configuration? Looks broken to me:

``` console
$ helm generate deis-dev
[ERROR] Error opening value file: Near line 45 (last key parsed 'gcs.key_json'): Expected value but found '`' instead.
[ERROR] Failed to complete generation: failed to execute helm template -o /Users/matt/.helm/workspace/charts/deis-dev/manifests/deis-registry-rc.yaml -d /Users/matt/.helm/workspace/charts/deis-dev/tpl/objectstorage.toml /Users/matt/.helm/workspace/charts/deis-dev/tpl/deis-registry-rc.yaml (/Users/matt/.helm/workspace/charts/deis-dev/tpl/deis-registry-rc.yaml): exit status 1
```

krancour:
@mboersma I tested this with the default (file system) and it worked for me.  I'd be happy to try again to see if I can repro your issue.

krancour:
I just realized I tested this in the time between some comments I'd made and subsequent force push that introduced the issue.  That explains it.

PR #127, opened 2016-03-01
feat(postgres): use object storage tempaltes

kmala:
use object storage template to populate the object storage details onto the container.

### DIFF
--- a/deis-dev/tpl/deis-objectstorage-secret.yaml
+++ b/deis-dev/tpl/deis-objectstorage-secret.yaml
@@ -17,4 +17,5 @@ data: {{ if eq .storage "gcs" }}
   accesskey: {{ b64enc .s3.accesskey }}
   secretkey: {{ b64enc .s3.secretkey }}
   region: {{ b64enc .s3.region }}
-  bucket: {{ b64enc .s3.bucket }}{{ end }}
+  bucket: {{ b64enc .s3.bucket }}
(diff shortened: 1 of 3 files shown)


mboersma:
Let's be careful not to merge this until the related code changes have been merged and the updated images actually make it to quay.io...

deis-dev/tpl/deis-objectstorage-secret.yaml
```diff
@@ -17,4 +17,5 @@ data: {{ if eq .storage "gcs" }}
   accesskey: {{ b64enc .s3.accesskey }}
   secretkey: {{ b64enc .s3.secretkey }}
   region: {{ b64enc .s3.region }}
-  bucket: {{ b64enc .s3.bucket }}{{ end }}
+  bucket: {{ b64enc .s3.bucket }}
+  database-bucket: {{ b64enc .s3.database-bucket }}{{ end }}
```
bacongobbler, line 21: what's the difference between bucket and database-bucket? Can't we store registry and the database's information in the same bucket through path separation?

kmala, line 21: Thought about it..but @mboersma has said that there may be few users who would want them in separate buckets

technosophos, line 21: Yeah, I kinda think it would be better to make it easy for an operator to store them separately. It's easy enough for them to set both to the same thing if they prefer that.

PR #131, opened 2016-03-03
fix(deis-dev): use minio as the default storage

kmala:
Use minio as the default storage backend. Add the accesskey and secret key of the minio to the object store secret so that we have one place where all the object store details are present. Once all the components use minio details from the object store we can delete the minio-user secret.

### DIFF
--- a/deis-dev/tpl/deis-objectstorage-secret.yaml
+++ b/deis-dev/tpl/deis-objectstorage-secret.yaml
@@ -8,7 +8,9 @@ metadata:
   annotations:
     deis.io/objectstorage: "{{.storage}}"
 type: Opaque
-data: {{ if eq .storage "gcs" }}
+data: {{ if eq .storage "minio" }}
+  access-key-id: OFRaUlkySlJXTVBUNlVNWFI2STUK
+  access-secret-key: Z2JzdHJPdm90TU1jZzJzTWZHVWhBNWE2RXQvRUk1QUx0SUhzb2JZawo={{ else if eq .storage "gcs" }}
   key.json: {{ b64enc .gcs.key_json }}
   bucket: {{ b64enc .gcs.bucket }}{{ else if eq .storage "azure" }}
   accountname: {{ b64enc .azure.accountname }}
--- a/deis-dev/tpl/deis-registry-rc.yaml
+++ b/deis-dev/tpl/deis-registry-rc.yaml
@@ -38,7 +38,7 @@ spec:
             - name: REGISTRY_LOG_LEVEL
               value: info
             - name: REGISTRY_STORAGE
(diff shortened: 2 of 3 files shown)


bacongobbler:
don't we need to change objectstorage.yaml to set minio as the default too, rather than filesystem?

kmala:
yes...that's one way but i want to remove the minio-user secret totally because then we need to mount both the secrets( minio and objectstore).

deis-dev/tpl/deis-objectstorage-secret.yaml
```diff
@@ -8,7 +8,9 @@ metadata:
   annotations:
     deis.io/objectstorage: "{{.storage}}"
 type: Opaque
-data: {{ if eq .storage "gcs" }}
+data: {{ if eq .storage "minio" }}
+  access-key-id: OFRaUlkySlJXTVBUNlVNWFI2STUK
```
bacongobbler, line 12: This was accidentally committed.

arschles:
@kmala I'm removing my `LGTM2` because of @bacongobbler's [comment](https://github.com/deis/charts/pull/131/files#r55242917). When that's resolved, please ping me and I'll re review.

deis-dev/tpl/deis-objectstorage-secret.yaml
kmala, line 12: No.This was on purpose because i think we need to use one secret for object store instead of two.Slowly we can remove the usage of minio-user secret from everywhere and remove the minio-user secret from the charts.

bacongobbler, line 12: We should generate this in objectstorage.yaml then in the same way minio-user is generated instead of hard-coding the access key and secret for every installation.

kmala, line 12: Right now minio-user is also hardcoded. We can change this from hard-coding to generating some random key and secret once all the repo uses the object-store secret instead of minio-user

arschles, line 12: @kmala is it possible to use the same generator to generate both keys? not necessary for this PR IMO, but it would be nice to have that. If it is possible, can you please create an issue for it?

kmala, line 12: Yes..that is my plan once i remove the usage of minio-secret from all repos.

kmala, line 12: @arschles can you LGTM if you are fine with this for now.

arschles:
@kmala I've re-added my `LGTM2`. This PR is ok to merge

PR #132, opened 2016-03-04
feat(storage): add storage envs to builder and workflow rcs

smothiki:
builder and workflow will support unified deis-object store model.
Not mounting deis-object store secret to workflow. As workflow doesn;t actually talk to object store but creates a object store secret for an APP.
We have to make changes to workflow to read secret from API server and populate it back in app namespace

### DIFF
--- a/deis-dev/manifests/deis-builder-rc.yaml
+++ b/deis-dev/manifests/deis-builder-rc.yaml
@@ -29,6 +29,8 @@ spec:
               value: "8092"
             - name: "EXTERNAL_PORT"
               value: "2223"
+            - name: BUILDER_STORAGE
+              value: {{default "minio" .storage}}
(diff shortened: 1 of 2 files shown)


smothiki:
@helgi @arschles ping

arschles:
@smothiki I think this is fine from the builder side of the house since it doesn't remove the secret and env vars that builder uses now.

Ref https://github.com/deis/builder/issues/224 for the work we need in builder to make this happen, right?

smothiki:
Yeah very much necessary. thanks for the LGTM guys. @helgi are you fine with reading deis-object store secret from API server rather than from filesystem ?

smothiki:
@arschles I want this https://github.com/deis/builder/pull/223 merged let's bug people till they LGTM

arschles:
It's merged

smothiki:
awesome

PR #133, opened 2016-03-04
feat(fix): move builder and workflow rc to tpl direcotry

smothiki:
moved templated rcs of builder and workflow to tpl directory


smothiki:
testing locally

PR #144, opened 2016-03-04
fix(deis-dev/tpl/deis-builder-rc.yaml): make the objectstore-creds secret read-only

arschles:
As far as I know, builder does not write to this volume, nor should it.

@smothiki thoughts?


smothiki:
Code LGTM . Yeah builder just reads the secret

jchauncey:
This needs a rebase but looks like we should have this change. Can we merge this?

arschles:
Not sure if we can merge, I haven't tested in a long time. Moving to RC1 with the intention of testing there

arschles:
Pushing this to v2, as testing is blocked by https://github.com/deis/builder/issues/351

PR #153, opened 2016-03-11
ref(deis-dev): rename workflow to controller

bacongobbler:
Workflow is the platform product; controller is the API server.

refs https://github.com/deis/deis/issues/4952

to be merged after https://github.com/deis/workflow/pull/494 goes green and has pushed the images to dockerhub/quay


arschles:
@bacongobbler looks like https://github.com/deis/workflow/pull/494 has gone green but needs merge before it'll push to the docker repos. LGTMing now, pending that of course.

PR #146, opened 2016-03-06
feat(dockerfile): Added Dockerfile for deis chart e2e test

chaitanyaenr:
Docker image for deis/charts e2e test

PR #161, opened 2016-03-15
chore(router): Transition to canary images where it makes sense

krancour:
Complements deis/router#146


arschles:
@krancour doesn't this depend on deis/router#146?

krancour:
Waiting for merge of #146 to produce its first canary image before I merge this.

PR #171, opened 2016-03-17
feat(storage): add details for the builder and database buckets

kmala:
add details for the builder and database buckets


workflow-dev/tpl/objectstorage.toml
```diff
@@ -29,16 +29,21 @@ secretkey = "YOUR SECRET HERE"
 # Any S3 region
 region = "us-west-1"
 # Your buckets.
-bucket = "your-bucket-name"
+bucket = "your-registry-bucket-name"
```
arschles: can we name this `registry_bucket` to remove ambiguity?

PR #175, opened 2016-03-17
chore(workflow-beta1-test): create chart with immutable git shas

sgoings:


workflow-beta1-test/Chart.yaml
```diff
@@ -0,0 +1,11 @@
+name: workflow-dev
```
bacongobbler, line 1: should be renamed to workflow-beta1-test

sgoings, line 1: Good call

jchauncey:
@sgoings merge it :ship:

workflow-beta1-test/Chart.yaml
sgoings, line 1: I'll add a pr to fix this once I'm back at home.

PR #179, opened 2016-03-18
chore(router): Cleanup standalone router charts

krancour:
Includes:
- `router` chart (alpha) goes away; gets replaced with `router-beta1-test`
- `router-beta1-test` manifests are copies of relevant manifests from `workflow-beta1-test`
- Manifest for `deis` namespace gets added to both `router-beta1-test` and `router-dev` charts.  These "standalone" router charts are mutually exclusive with installing all of Workflow-- so these charts must account for namespace creation themselves.  It was an oversight that they did not previous and it went undetected because some clusters automatically create namespaces, depending on what admission controls are in place.


router-beta1-test/Chart.yaml
```diff
@@ -1,6 +1,6 @@
-name: router
+name: router-beta1-test
 home: https://github.com/deis/router
-version: 2.0.0-alpha
+version: v2-beta
```
arschles, line 3: are we calling this beta `v2-beta` across the board?

router-beta1-test/README.md
```diff
@@ -1,4 +1,4 @@
-# Deis Router 2.0.0-alpha
+# Deis Router 2.0.0-beta
 
 WARNING: this chart is for testing only! Features may not work and there are likely to be bugs.
```
arschles, line 3: maybe change to indicate beta status, instead of testing only?

router-beta1-test/Chart.yaml
krancour, line 3: Good question.  I just emulated what the `workflow-beta1-test` chart had in this same field, with the assumption that it was correct.

router-beta1-test/README.md
krancour, line 3: Again... I just followed the pattern from the `workflow-beta1-test` chart with the assumption that it was correct.

router-beta1-test/Chart.yaml
arschles, line 3: K, sounds like there's consistency, so I'm good with this

router-beta1-test/README.md
arschles, line 3: Same here regarding consistency.

jchauncey, line 3: So I made logger v2.0.0-beta1 just FYI

krancour:
@arschles thanks for the review.  fwiw... I think your questions are broadly applicable to the multiple `*-beta1-test` charts we have atm... perhaps it's worth opening a serparate issue to get some clarification on it and/or track the fix if needed.

arschles:
@krancour good idea - created https://github.com/deis/charts/issues/184 for naming consistency

PR #180, opened 2016-03-18
chore(logger): Add beta1 chart for logger

jchauncey:


deis-logger-beta1-test/Chart.yaml
```diff
@@ -0,0 +1,12 @@
+name: deis-logger-beta1-test
+home: http://github.com/deis/deis
+version: 0.1.0
```
krancour: I would change that.  Looks like the `workflow-beta1-test` chart uses a value of `v2-beta` here.

deis-logger-beta1-test/manifests/deis-logger-fluentd-daemon.yaml
```diff
@@ -0,0 +1,39 @@
...
+    metadata:
+      name: deis-logger-fluentd
+      labels:
+        heritage: deis
+    spec:
+      serviceAccount: deis-logger-fluentd
+      containers:
+      - name: deis-logger-fluentd
+        image: quay.io/deisci/fluentd:git-f537de4
+        imagePullPolicy: Always
```
krancour: If we're pulling an immutable tag, this should change to `IfNotPresent`.  Looks like `workflow-beta1-test` has that wrong in a couple spots (but right in most spots).  I'll PR that separately, but this should be fixed here.

deis-logger-beta1-test/manifests/deis-logger-rc.yaml
```diff
@@ -0,0 +1,38 @@
...
+    app: deis-logger
+  template:
+    metadata:
+      labels:
+        app: deis-logger
+    spec:
+      containers:
+      - name: deis-logger
+        image: quay.io/deisci/logger:git-a16379c
+        imagePullPolicy: Always
```
krancour: Again, use `imagePullPolicy: IfNotPresent`

deis-logger-beta1-test/Chart.yaml
jchauncey: i would rather this be v2.0.0-beta1 if we are going to use an actual version here. @sgoings @slack @mboersma opinions?

mboersma: I think `v2.0.0-beta1` is more correct and useful than `v2-beta`.

jchauncey:
ok updated PR

deis-logger-beta1-test/manifests/deis-logger-svc.yaml
```diff
@@ -0,0 +1,19 @@
...
+  ports:
+  - port: 80
+    name: http
+    targetPort: http
+  - port: 514
+    name: transport
+    targetPort: 1514
+    protocol: UDP
+  selector:
+    app: deis-logger
```
arschles, line 19: Are we putting newlines at the end of files? I prefer yes, but no strong feelings. This PR is heterogeneous though.

jchauncey, line 19: Some people hate them others want them. I don't even notice them to be
honest.

arschles:
@jchauncey this is ok to ship. My comment above (https://github.com/deis/charts/pull/180#discussion_r56727887) is not blocking for this PR

deis-logger-beta1-test/manifests/deis-logger-svc.yaml
mboersma, line 19: :+1: to including newlines at the ends of files by default. Some tools will complain about the lack of a final newline, but none will complain about it being there.

PR #186, opened 2016-03-22
feat(deis-tests): mount /tmp emptyDir for storing artifacts

bacongobbler:
Allows us to store build artifacts in the sidecar container for future processing after the tests have finished running. I decided on `/root` because that is the home directory when running the tests.

Example:

```
><> kd exec -c deis-e2e deis-tests -- sh -c 'echo "hello" > /root/foo'
><> kd exec -c artifacts deis-tests -- cat /root/foo
hello
```

BREAKING CHANGES:

with `kubectl --namespace=deis logs`, you now have to specify the container name with `-c`. Because of this, changes to sgoings/chart-mate also need to occur so that it will continue to work with this change.

History and motivation behind this change is for https://github.com/sgoings/chart-mate/pull/28

### DIFF
--- a/deis-tests/manifests/deis-tests-pod.yaml
+++ b/deis-tests/manifests/deis-tests-pod.yaml
@@ -6,6 +6,7 @@ metadata:
   labels:
     heritage: deis
 spec:
+  restartPolicy: Never
   containers:
   - name: deis-e2e
     image: quay.io/deisci/deis-e2e:canary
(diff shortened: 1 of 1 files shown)


bacongobbler:
I thought I was hitting restart policy errors before but everything seems to be fine now after manually testing. This should be ready for review now along with deis/workflow-e2e#120 and sgoings/chart-mate#28.

vdice:
LGTM.  Tested locally (paired w/ change in https://github.com/sgoings/chart-mate/pull/28) and junit.xml artifact produced as expected:

```
$ ls -haltr ~/.chart-mate/logs/junit.xml
-rw-r--r--  1 vaughndice  staff    18K Mar 23 13:37 /Users/vaughndice/.chart-mate/logs/junit.xml
```

PR #193, opened 2016-03-25
fix(component_jobs.groovy) remove extra brace that snuck in with last PR

vdice:
(This is same change as was approved in https://github.com/deis/charts/pull/190)


vdice:
cc @sgoings @jchauncey ONE LAST TIME :grin:

sgoings:
How's that for an LGTM? ;-)

PR #197, opened 2016-03-25
feat(workflow_test.groovy): archive (log) artifacts

vdice:
inspiration/settings taken from @jchauncey's setup in https://ci.deis.io/job/workflow-beta1-test-e2e/

cc @jchauncey @sgoings


vdice:
tested example can be seen here https://ci.deis.io/job/workflow-test-pr-clone/

PR #200, opened 2016-03-28
feat(settings): use k8s 1.2 secrets as env vars to configure production settings

helgi:
Keeping the mount points for now since they do not harm anything

http://kubernetes.io/docs/user-guide/secrets/#using-secrets-as-environment-variables


krancour:
Can't the volume mounts for the secrets go away now?

PR #201, opened 2016-03-30
feature(deis-tests): Add a chart that runs the workflow-e2e tests in parallel

jchauncey:
This PR also changes how we launch the sequential workflow-e2e tests

this pr is required for https://github.com/deis/workflow-e2e/pull/135

### DIFF
--- a/deis-tests-parallel/manifests/deis-tests-parallel-pod.yaml
+++ b/deis-tests-parallel/manifests/deis-tests-parallel-pod.yaml
@@ -12,14 +12,10 @@ spec:
     image: quay.io/deisci/deis-e2e:canary
     imagePullPolicy: Always
     command: ["ginkgo"]
-    args: ["-slowSpecThreshold=120.00", "-noisyPendings=false", "-progress", "tests/"]
+    args: ["-slowSpecThreshold=120.00", "-nodes=30", "-noisyPendings=false", "-progress", "tests/"]
     env:
       - name: JUNIT
         value: "true"
-      - name: DEFAULT_MAX_TIMEOUT
-        value: "300s"
-      - name: DEFAULT_EVENTUALLY_TIMEOUT
-        value: "30s"
     volumeMounts:
     - name: artifact-volume
       mountPath: /root
--- a/deis-tests-parallel/Chart.yaml
+++ b/deis-tests-parallel/Chart.yaml
@@ -1,0 +1,12 @@
+name: deis-tests-parallel
+home: https://github.com/deis/workflow-e2e
+version: v2-beta
+description: End-to-end tests for Deis v2 which are run in parallel
+maintainers:
+- Deis Team <engineering@deis.com>
+details: |-
+  Runs the End-to-end (e2e) workflow tests for the Deis v2 open source PaaS in parallel.
+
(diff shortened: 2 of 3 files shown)


deis-tests-parallel/manifests/deis-tests-parallel-pod.yaml
```diff
@@ -0,0 +1,31 @@
...
+    args: ["-slowSpecThreshold=120.00", "-nodes=30", "-noisyPendings=false", "-progress", "tests/"]
+    env:
+      - name: JUNIT
+        value: "true"
+    volumeMounts:
+    - name: artifact-volume
+      mountPath: /root
+  - name: artifacts
+    image: busybox
+    imagePullPolicy: Always
```
arschles, line 24: need to always pull? seems like since you're just running `tail`, it's not a big deal if you have something outdated. even better would be to specify a specific tag that's not `latest`

jchauncey, line 24: it was set to always in the other deis-tests chart so it got copied here. i agree that i think these tests should probably be pinned to a specific version

arschles, line 24: Ok, my vote is for pinning to a specific version and specifying `IfNotPresent` for the pull policy. Since `tail` is probably gonna be available on most busybox versions going forward and it's not a big deal to pull such a (relatively) small image, I wouldn't say it's blocking for this PR. Mind creating an issue in this repo to at least investigate making the change though?

deis-tests-parallel/manifests/deis-tests-parallel-pod.yaml
```diff
@@ -0,0 +1,31 @@
...
+  labels:
+    heritage: deis
+spec:
+  restartPolicy: Never
+  containers:
+  - name: deis-e2e
+    image: quay.io/deisci/deis-e2e:canary
+    imagePullPolicy: Always
+    command: ["ginkgo"]
+    args: ["-slowSpecThreshold=120.00", "-nodes=30", "-noisyPendings=false", "-progress", "tests/"]
```
kmala, line 15: did you try using "-stream" option..It's really helpful when running in parallel.

```
Setting stream to true will, instead, stream output from all parallel nodes in real-time, prepending each log line with the node # that emitted it
```

arschles, line 15: wouldn't it make more sense to put these args into the workflow-e2e dockerfile, so that it isn't so tightly coupled to this repo?

jchauncey, line 15: @vdice  the `-stream` option IIRC made the output really noisy. Instead by turning off the verbose mode and using `GinkgoWriter` it will output the correct log messages for tests failures and nothing else. 

@arschles yes except that I figured we might want to make these configurable for people who are running them on their own cluster. I didnt want to bake all of that into the image.

arschles, line 15: if you put all of this into a `CMD` in the `Dockerfile`, wouldn't we get both the loose coupling and configurability?

PR #209, opened 2016-04-01
fix(*): Jenkins sets SUDO_USER which messes with rerun behavior now

sgoings:


ci.sh
```diff
@@ -4,12 +4,11 @@ set -eo pipefail
 
 rm -rf ${HOME}/.chart-mate
 
-export CI=true
-
```
sgoings, line 8: There's no more usage of this variable since I split out bumpver into its own command.

mboersma:
:applause: for figuring this out. I wasted a few hours yesterday banging my head against this issue on node8-ec2 and node9-ec2.

PR #211, opened 2016-04-01
chore(deis-tests): Make deis-tests use old binary to run test suite

jchauncey:
This doesnt actually revert any of the commits but it takes the last known good deis-tests-pod manifest and puts it back.

### DIFF
--- a/deis-tests/manifests/deis-tests-pod.yaml
+++ b/deis-tests/manifests/deis-tests-pod.yaml
@@ -9,17 +9,13 @@ spec:
   restartPolicy: Never
   containers:
   - name: deis-e2e
-    image: quay.io/deisci/deis-e2e:canary
+    image: quay.io/deisci/deis-e2e:git-7f7d775
     imagePullPolicy: Always
-    command: ["ginkgo"]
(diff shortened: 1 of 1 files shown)


deis-tests/manifests/deis-tests-pod.yaml
```diff
@@ -9,17 +9,13 @@ spec:
   restartPolicy: Never
   containers:
   - name: deis-e2e
-    image: quay.io/deisci/deis-e2e:canary
+    image: quay.io/deisci/deis-e2e:git-7f7d775
```
mboersma, line 12: Shouldn't this chart always be referencing the `:canary` tag though? If the tests themselves are broken, we should fix the tests so that we get a new canary image. Sorry, I must be missing some detail here.

jchauncey, line 12: i thought it used canary too. but at some point over the last day or 2 we moved to using immutable images in the chart.

mboersma, line 12: > we moved to using immutable images in the chart.

Wow, you snooze, you lose around here. :open_mouth: That's even better then.

PR #212, opened 2016-04-05
chore(deis-tests-parallel): Back down the number of concurrent test processes

jchauncey:


vdice:
@jchauncey It could be handy as we go thru the trial-and-error process of tuning the parallel test runs if this particular value was configurable via an env var? Then we can have a default and be able to config on the fly (via the Jenkins job.)

Aside from that (we can create a ticket if it sounds helpful), this LGTM.

jchauncey:
yeah you know how we could do that?

vdice:
Not immediately, but figure there are examples of other component manifests that use env vars in a similar fashion.

jchauncey:
lets go ahead and back it down so i can watch the parallel runs today.

PR #214, opened 2016-04-08
ref(workflow-e2e): Use default command specified by image

krancour:
This change complements https://github.com/deis/workflow-e2e/pull/157

The idea here is that the e2e test image itself will specify the default command used to kick off the tests.  This consolidates responsibility for _how_ the tests are executed into the deis/workflow-e2e repo and reduces the need for future changes to this chart if / as _how_ we execute tests changes.

At the same time, I've eliminated setting the `JUNIT` env var since that is defaulted to true in the image.

cc @vdice

### DIFF
--- a/workflow-beta1-e2e/manifests/deis-tests-pod.yaml
+++ b/workflow-beta1-e2e/manifests/deis-tests-pod.yaml
@@ -11,11 +11,6 @@ spec:
   - name: deis-e2e
     image: quay.io/deis/deis-e2e:v2.0.0-beta1
     imagePullPolicy: IfNotPresent
-    command: ["/bin/tests.test"]
-    args: ["-test.v", "-test.timeout=60m", "-ginkgo.v", "-ginkgo.slowSpecThreshold=120"]
-    env:
-      - name: JUNIT
-        value: "true"
     volumeMounts:
     - name: artifact-volume
       mountPath: /root


vdice:
@krancour I believe we want to leave the `workflow-beta1-e2e` chart 'immutable' (as it is tied to the `quay.io/deis/deis-e2e:v2.0.0-beta1` image).  Rather, I propose this change be made in the `deis-tests(-parallel)` charts [here](https://github.com/deis/charts/tree/master/deis-tests) and [here](https://github.com/deis/charts/tree/master/deis-tests-parallel).

jchauncey:
what you submitted is out of date with what is currently in master https://github.com/deis/charts/blob/master/deis-tests/manifests/deis-tests-pod.yaml

jchauncey:
basically compiling a test binary doesn't allow us to run the tests in parallel so we have to use ginkgo

krancour:
@vdice... yup.  This was a mistake.  You are correct that I modified the wrong chart.  (That being said `deis-tests` seems poorly name, imho.  These _are_ specifically Workflow tests.)

First thing on Monday, I'll PR the correct chart. (On my way out the door for the weekend, currently.)

@jchauncey ya.  I know.  Don't worry.  The `CMD` directive on the image looks like this, which is fine.  It does not use statically compiled tests, but rather uses `make` to execute `ginkgo`

```
CMD ["/usr/bin/make", "test-integration"]
```

krancour:
Removing awaiting review tag since some updates to this PR are required.

krancour:
Closing in favor of #218

PR #219, opened 2016-04-11
feat(logger): Add logger to the workflow-dev chart

jchauncey:
Now that we support 1.2 we can add it as part of the default deis install

### DIFF
--- a/workflow-dev/manifests/deis-logger-fluentd-daemon.yaml
+++ b/workflow-dev/manifests/deis-logger-fluentd-daemon.yaml
@@ -1,0 +1,39 @@
+apiVersion: extensions/v1beta1
+kind: DaemonSet
+metadata:
+  name: deis-logger-fluentd
+  namespace: deis
+  labels:
+    heritage: deis
+spec:
+  template:
+    metadata:
+      name: deis-logger-fluentd
+      labels:
+        heritage: deis
+    spec:
+      serviceAccount: deis-logger-fluentd
+      containers:
+      - name: deis-logger-fluentd
+        image: quay.io/deisci/fluentd:canary
+        imagePullPolicy: Always
+        securityContext:
+          privileged: true
+        volumeMounts:
+        - name: varlog
+          mountPath: /var/log
+        - name: varlibdockercontainers
+          mountPath: /var/lib/docker/containers
+          readOnly: true
+        env:
+        - name: SYSLOG_HOST
+          value: $(DEIS_LOGGER_SERVICE_HOST)
+        - name: SYSLOG_PORT
+          value: $(DEIS_LOGGER_SERVICE_PORT_TRANSPORT)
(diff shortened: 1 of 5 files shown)


workflow-dev/manifests/deis-logger-fluentd-daemon.yaml
```diff
@@ -0,0 +1,39 @@
...
+    metadata:
+      name: deis-logger-fluentd
+      labels:
+        heritage: deis
+    spec:
+      serviceAccount: deis-logger-fluentd
+      containers:
+      - name: deis-logger-fluentd
+        image: quay.io/deisci/fluentd:canary
+        imagePullPolicy: IfNotPresent
```
arschles: IMO we should use `Always` since this is in `workflow-dev`

workflow-dev/manifests/deis-logger-fluentd-daemon.yaml
```diff
@@ -0,0 +1,39 @@
...
+          value: $(DEIS_LOGGER_SERVICE_HOST)
+        - name: SYSLOG_PORT
+          value: $(DEIS_LOGGER_SERVICE_PORT_TRANSPORT)
+      volumes:
+      - name: varlog
+        hostPath:
+          path: /var/log
+      - name: varlibdockercontainers
+        hostPath:
+          path: /var/lib/docker/containers
```
arschles, line 39: is this gonna be uniform on all Kubernetes installs?

workflow-dev/manifests/deis-logger-fluentd-daemon.yaml
arschles: same in below `imagePullPolicy` values

workflow-dev/manifests/deis-logger-svc.yaml
```diff
@@ -0,0 +1,19 @@
...
+  ports:
+  - port: 80
+    name: http
+    targetPort: http
+  - port: 514
+    name: transport
+    targetPort: 1514
+    protocol: UDP
+  selector:
+    app: deis-logger
```
arschles: can you add a newline here for uniformity across files?

workflow-dev/manifests/deis-logger-fluentd-daemon.yaml
jchauncey: good catch this was a copy from the beta1 chart =)

workflow-dev/manifests/deis-logger-fluentd-daemon.yaml
jchauncey, line 39: So it seems that right now its pretty uniform but we are finding edge cases that im going to address in another PR (i need to fix it in fluentd first)

jchauncey:
@arschles updated

PR #220, opened 2016-04-11
clean(minio): remove unused secrets and secret references

kmala:


helgi:
You need to remove `manifest/deis-minio-secret-user.yaml` as well I believe

kmala:
I can't remove that for now...To remove it i need to change the minio code and also i need to change the charts so that minio rc gets created only when storage type is minio. I had made changes for that separately and will create a PR once i finish testing.

PR #222, opened 2016-04-13
bug(labels): create daemonset with pod selectors

slack:


slack:
I need to do this against `workflow-dev` as well, juan momento.

slack:
deis-logger needs an update as well?

bacongobbler:
manually tested; works. the deis-workflow-e2e pod is no longer destroyed.

bacongobbler:
Wouldn't hurt to do so

vdice:
Also tested manually -- tests are now running (`workflow-dev-e2e` not deleted by daemonset)

PR #231, opened 2016-04-16
style(builder): disable debug output in beta2 chart

slack:


mboersma:
See also deis/dockerbuilder#46. tl:dr do this to `deis-builder-rc.yaml` to also silence dockerbuilder spew:

``` yaml
- name: "DOCKERBUILDER_IMAGE_NAME"
  value: "quay.io/deisci/dockerbulder:git-8715439"
- name: "DEBUG"
  value: "false"
```

PR #235, opened 2016-04-18
chore: Allows for specifying the docker tag of a component image as an environment variable

jchauncey:
This changes only the workflow-dev and workflow-dev-e2e charts for now.

### DIFF
--- a/workflow-dev-e2e/tpl/workflow-e2e-pod.yaml
+++ b/workflow-dev-e2e/tpl/workflow-e2e-pod.yaml
@@ -10,7 +10,7 @@ spec:
   restartPolicy: Never
   containers:
   - name: tests
-    image: quay.io/{{.e2e.org}}/workflow-e2e:{{.e2e.dockerTag}}
+    image: quay.io/{{.e2e.org}}/workflow-e2e:{{env WORKFLOW_E2E_GIT_TAG | default .e2e.dockerTag}}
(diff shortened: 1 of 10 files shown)


workflow-dev/manifests/deis-logger-fluentd-daemon.yaml
```diff
@@ -20,7 +21,7 @@ spec:
       serviceAccount: deis-logger-fluentd
       containers:
       - name: deis-logger-fluentd
-        image: quay.io/deisci/fluentd:canary
+        image: quay.io/deisci/fluentd:208d0dc
         imagePullPolicy: Always
         securityContext:
           privileged: true
```
arschles, line 27: can you remove this file and ensure that its templated version is in `workflow-dev/tpl`?

jchauncey, line 27: yup forgot to remove it. should be updated now

jchauncey:
This PR also needs https://github.com/sgoings/chart-mate/pull/58

jchauncey:
So heres the plan. Instead of us using the bump command in chart-mate which does a sed replace of the values in the chart. We will now use the environment variables specified in each template. 

The corresponding change in chart-mate sets each `COMPONENT_GIT_TAG` value to be `"git-${COMPONENT_SHA:0:7}"`

PR #237, opened 2016-04-19
feat(workflow-dev): store builder ssh private keys in secret

bacongobbler:
PR in tandem with deis/builder#302


bacongobbler:
relies on helm 0.6.0, which is pending release after https://github.com/helm/helm/pull/440.

PR #243, opened 2016-04-26
ref(termination): for dev / e2e only make the termination grace period 5s instead of 30s

helgi:


vdice:
code LGTM; will run tests off this change...

workflow-dev/tpl/deis-controller-rc.yaml
```diff
@@ -36,6 +36,8 @@ spec:
             - containerPort: 8000
               name: http
           env:
+            - name: KUBERNETES_POD_TERMINATION_GRACE_PERIOD_SECONDS
+              value: 5
```
vdice: When running locally, I get the following on installation of chart: 

`unable to decode "STDIN": [pos 860]: json: expect char '"' but got char '5'`

Thinking this needs to read `value: "5"`

vdice:
Tests looking good and ~3 mins or so faster:

```
Ran 152 of 155 Specs in 471.722 seconds
SUCCESS! -- 152 Passed | 0 Failed | 3 Pending | 0 Skipped

Ginkgo ran 1 suite in 7m57.58422911s
Test Suite Passed
```

PR #252, opened 2016-05-05
ref(*): rename workflow-dev components from deis-* to *

### LINKED ISSUE #155: rename deis-dev components from deis-* to *
This seems ugly and redundant:

```
kubectl --namespace=deis get svc deis-router
```

If it serves no purpose then I propose that we rename service names to their bare names, i.e.

```
kubectl --namespace=deis get svc router
```

I understand that this is a holistic change (since it affects dns and the pod's environment) and probably shouldn't be done pre-beta but we should consider it.

helgi:
fixes #155


krancour:
The router will require a complementary PR.  Router is constantly looking for its own RC _by name_.

helgi:
Should I also rename all the files??

helgi:
I moved the files to ditch the prefix

bacongobbler:
multiple components are going to need changes, since the service names have changed. For example the database looks for DEIS_MINIO_SERVICE_HOST, which we'd need to change to MINIO_SERVICE_PORT. Other than that I'm happy with this change. :+1:

helgi:
That's a good point - I forgot those are auto generated based on that. Too bad it isn't `namespace + service name` kind of dealio but we'd have to fix that anyway...

I'll look at how much work it is to change all those things and how many projects are affected - Maybe not a beta4 timeframe anymore

mboersma:
This PR hasn't been touched in over five months, and Workflow is about to prefer [kubernetes/helm](https://github.com/kubernetes/helm) for installation. I'm closing this, but please re-open if it's still worthy of consideration.

PR #253, opened 2016-05-06
Added k8s-home URL to README.md

nicdoye:
Trivial change. Link was missing.

For consistency, we could:
- change all the remote URLs to be at the bottom, like these 
- change the kubernetes one, I've just added to be an in-line link, like all the others.


bacongobbler:
thanks @nicdoye!

PR #261, opened 2016-05-10
docs(workflow-dev chart): update README

vdice:


workflow-dev/README.md
```diff
@@ -1,15 +1,17 @@
-# Workflow 2.0.0-beta
+# Workflow 2.0.0-dev
```
vdice, line 1: this string change will enable `deisrel` to swap out `dev` for `$WORKFLOW_RELEASE_SHORT` when creating a release chart...

PR #266, opened 2016-05-17
feat(tpl): template object-store-creds to take environment variables

smothiki:
stuck with this bug 

```
region = "us-west-1"
# Your buckets.
registry_bucket = "[ERROR] Template rendering failed: template: helmTpl:33:46: executing "helmTpl" at <randAlpha>: wrong number of args for randAlpha: want 1 got 1
```

### DIFF
--- a/workflow-dev/tpl/deis-objectstorage-secret.yaml
+++ b/workflow-dev/tpl/deis-objectstorage-secret.yaml
@@ -7,23 +7,23 @@ metadata:
   labels:
     heritage: deis
   annotations:
-    deis.io/objectstorage: "{{.storage}}"
+    deis.io/objectstorage: "{{ or (env "STORAGE_TYPE") (.storage)}}"
 type: Opaque
-data: {{ if eq .storage "minio" }}
+data: {{ if or (eq (env "STORAGE_TYPE") "gcs") (eq .storage "gcs")}}
+  key.json: {{ or (env "GCS_KEY_JSON") (.gcs.key_json) | b64enc}}
+  registry-bucket: {{ or (env "GCS_REGISTRY_BUCKET") (.gcs.registry_bucket) | b64enc}}
+  database-bucket: {{ or (env "GCS_DATABASE_BUCKET") (.gcs.registry_bucket) | b64enc }}
+  builder-bucket: {{ or (env "GCS_BUILDER_BUCKET") (.gcs.builder_bucket) | b64enc }}{{ else if or (eq (env "STORAGE_TYPE") "azure") (eq .storage "azure")}}
+  accountname: {{ or (env "AZURE_ACCOUNT_NAME") (.azure.accountname) | b64enc }}
+  accountkey: {{ or (env "AZURE_ACCOUNT_KEY") (.azure.accountkey) | b64enc }}
+  registry-container: {{ or (env "AZURE_REGISTRY_CONTAINER") (.azure.registry_container) | b64enc }}
+  database-container: {{ or (env "AZURE_DATABASE_CONTAINER") (.azure.database_container) | b64enc }}
+  builder-container: {{ or (env "AZURE_BUILDER_CONTAINER") (.azure.builder_container) | b64enc }}{{ else if or (eq (env "STORAGE_TYPE") "s3") (eq .storage "s3")}}
+  accesskey: {{ or (env "AWS_ACCESS_KEY") (.s3.accesskey) | b64enc }}
+  secretkey: {{ or (env "AWS_SECRET_KEY") (.s3.secretkey) | b64enc }}
+  region: {{ or (env "S3_REGION") (.s3.region) | b64enc }}
+  registry-bucket: {{ or (env "AWS_REGISTRY_BUCKET") (.s3.registry_bucket) | b64enc }}
+  database-bucket: {{ or (env "AWS_DATABASE_BUCKET") (.s3.database_bucket) | b64enc }}
+  builder-bucket: {{ or (env "AWS_DATABASE_BUCKET") (.s3.builder_bucket) | b64enc }}{{else if or (eq (env "STORAGE_TYPE") "minio") (eq .storage "minio")}}
   accesskey: OFRaUlkySlJXTVBUNlVNWFI2STU=
-  secretkey: Z2JzdHJPdm90TU1jZzJzTWZHVWhBNWE2RXQvRUk1QUx0SUhzb2JZaw=={{ else if eq .storage "gcs" }}
-  key.json: {{ b64enc .gcs.key_json }}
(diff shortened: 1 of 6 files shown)


workflow-dev/tpl/generate_params.toml
```diff
@@ -17,37 +17,37 @@
...
 # Any S3 region
-region = "us-west-1"
+region = "{{ env "REGION" | default "us-west-1" }}"
 # Your buckets.
-registry_bucket = "your-registry-bucket-name"
-database_bucket = "your-database-bucket-name"
-builder_bucket = "your-builder-bucket-name"
+registry_bucket = "{{ env "REGISTRY_BUCKET" | randAlpha 5 }}"
+database_bucket = "{{ env "DATABASE_BUCKET" | randAlpha 5 }}"
+builder_bucket = "{{ env "BUILDER_BUCKET" | randAlpha 5 }}"
```
arschles: since you prefixed the `accessKey` and `secretKey` env vars with `AWS_`, can you prefix these also with `AWS_`?

workflow-dev/tpl/generate_params.toml
```diff
@@ -17,37 +17,37 @@
...
-accountname = "YOUR ACCOUNT NAME"
-accountkey = "YOUR ACCOUNT KEY"
-registry_container = "your-registry-container-name"
-database_container = "your-database-container-name"
-builder_container = "your-builder-container-name"
+accountname = "{{ env "ACCOUNT_NAME" }}"
+accountkey = "{{ env "ACCOUNT_KEY" }}"
+registry_container = "{{ env "REGISTRY_BUCKET" | randAlpha 10 }}"
+database_container = "{{ env "DATABASE_BUCKET" | randAlpha 10 }}"
+builder_container = "{{ env "BUILDER_BUCKET" | randAlpha 10 }}"
```
arschles: since you prefixed the s3 env vars with `AWS_`, maybe it'd be a good idea to prefix these with `AZURE_` and the below with `GCS_`. Thoughts?

smothiki:
Now Object storage creds will take environment variables if not set fall back to generate_params.toml

jchauncey:
looks good. it could probably be cleaned up a little but LGTM

mboersma:
So now these env vars will override what the user puts in `generate_params.toml`? I think that's reasonable but it's worth asking if the opposite behavior would be preferable.

This needs a companion PR in deis/workflow to document how to use these environment variables to configure object storage.

smothiki:
I thought about the other way making generate_params.toml top priority but thing with that is every field has to be left blank, none of the files should say `"your test goes here"`. So I gave envs more priority than generate_params.toml.

smothiki:
Apologies this PR needs some other templates to be changed like builder , registry , controller , database which sets <>_STORAGE env from generate_params.toml . Pushing again

smothiki:
requires https://github.com/deis/workflow/pull/240

workflow-dev/tpl/deis-objectstorage-secret.yaml
```diff
@@ -1,4 +1,4 @@
-#helm:generate helm template -o $HELM_GENERATE_DIR/manifests/deis-objectstorage-secret.yaml -d $HELM_GENERATE_DIR/tpl/generate_params.toml $HELM_GENERATE_DIR/tpl/deis-objectstorage-secret.yaml
+#helm:generate helmc template -o $HELM_GENERATE_DIR/manifests/deis-objectstorage-secret.yaml -d $HELM_GENERATE_DIR/tpl/generate_params.toml $HELM_GENERATE_DIR/tpl/deis-objectstorage-secret.yaml
```
mboersma: I think this header should continue to reference `helm`, not `helmc`. That's what we have in the workflow-beta4 chart and I believe Helm Classic intends to keep it that way for backward compatibility. ping @krancour.

workflow-dev/tpl/minio.sh
```diff
@@ -1,12 +1,12 @@
...
-      rm -rf $HELM_GENERATE_DIR/manifests/deis-minio-*
-    fi
+store=`grep "storage = " $HELM_GENERATE_DIR/tpl/generate_params.toml |cut -d "=" -f2 | tr -dc '[:alnum:]\n\r'`
+if [ -n "$STORAGE_TYPE" ]; then
+  if [ "$STORAGE_TYPE" != "minio" ]; then
+    rm -rf $HELM_GENERATE_DIR/manifests/deis-minio-*
   fi
-done < "$HELM_GENERATE_DIR/tpl/generate_params.toml"
+elif [ "$store" != "minio" ]; then
+  rm -rf $HELM_GENERATE_DIR/manifests/deis-minio-*
```
vdice, line 8: Why do we need to actively destroy this file?

Assuming this logic is needed, is it possible to collapse the two conditions into one statement?  in pseudocode: `if $STORAGE_TYPE != "minio" OR $store != "minio", remove the correct file`

workflow-dev/tpl/deis-objectstorage-secret.yaml
```diff
@@ -7,23 +7,23 @@ metadata:
...
-  database-container: {{ b64enc .azure.database_container }}
-  builder-container: {{ b64enc .azure.builder_container }}{{ else if eq .storage "s3" }}
-  accesskey: {{ b64enc .s3.accesskey }}
-  secretkey: {{ b64enc .s3.secretkey }}
-  region: {{ b64enc .s3.region }}
-  registry-bucket: {{ b64enc .s3.registry_bucket }}
-  database-bucket: {{ b64enc .s3.database_bucket }}
-  builder-bucket: {{ b64enc .s3.builder_bucket }}{{ end }}
+data: {{ if or (eq (env "STORAGE_TYPE") "gcs") (eq .storage "gcs")}}
+key.json: {{ or (env "GCS_KEY_JSON") (.gcs.key_json) | b64enc}}
```
vdice: there's a lot going on here and it looks the previous indentation and spacing got removed -- any way we can add back some indentation/spacing for readability?

workflow-dev/tpl/minio.sh
smothiki, line 8: Problem with above logic is it get's passed if STORAGE_TYPE env is not set. So first we have to check if the variable is set and then look for STORAGE_TYPE != minio. deis-minio-\* files are present in tpl directory in manifests only tpl generated files are present, which are unnecessary if STORAGE_TYPE is not minio which adds an extra container, service, secret to the cluster.

smothiki:
@vdice corrected indentation

vdice:
thanks, @smothiki ... code LGTM but want to ping @mboersma since he had checked this PR previously and might want to review before merge.

Issue #267, opened 2016-05-20
Fix environment variables for the deis-workflow-manager pod


arschles:
The below environment variables should be set for the `deis-workflow-manager` pod. Some may be already set correctly. This issue is for ensuring that they _all_ are set properly:
- `PORT`
- `POLL_INTERVAL_SEC`
- `VERSIONS_API_URL`
- `DOCTOR_API_URL`
- `API_VERSION`
- `CHECK_VERSIONS`
- `DEIS_NAMESPACE`

PR #268, opened 2016-05-20
fix(workflow-manager): correct environment variables

arschles:
as of https://github.com/deis/workflow-manager/pull/56, these are the correct environment variables (some were changed before that pull request, but deis/workflow-manager#56 is the point at which all are guaranteed to be changed to these names)

Fixes https://github.com/deis/charts/issues/267

### DIFF
--- a/workflow-dev/tpl/deis-workflow-manager-rc.yaml
+++ b/workflow-dev/tpl/deis-workflow-manager-rc.yaml
@@ -30,7 +30,7 @@ spec:
         - name: VERSIONS_API_URL
           value: {{.workflowManager.versionsApiURL}}
         - name: DOCTOR_API_URL
-          value: "https://doctor.deis.com"
+          value: {{.workflowManager.doctorApiURL}}
         - name: POLL_INTERVAL_SEC
           value: "43200"
         - name: CHECK_VERSIONS
--- a/workflow-dev/tpl/deis-workflow-manager-rc.yaml
+++ b/workflow-dev/tpl/deis-workflow-manager-rc.yaml
@@ -25,19 +25,19 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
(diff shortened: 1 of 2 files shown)


bacongobbler:
LGTM3. I'll reflect this in deis/workflow#253

EDIT: done :)

arschles:
requires https://github.com/deis/workflow-manager/pull/56 before merge

workflow-dev/tpl/deis-workflow-manager-rc.yaml
```diff
@@ -25,19 +25,19 @@ spec:
...
             fieldRef:
               fieldPath: metadata.namespace
-        - name: WORKFLOW_MANAGER_PORT
+        - name: PORT
           value: "8080"
-        - name: WORKFLOW_MANAGER_VERSIONSAPIURL
+        - name: VERSIONS_API_URL
           value: {{.workflowManager.versionsApiURL}}
-        - name: WORKFLOW_MANAGER_DOCTORAPIURL
+        - name: DOCTOR_API_URL
```
vdice, line 32: wondering if we want to follow a similar pattern as `VERSIONS_API_URL` above?  If so, we can allow release charts go out with the prod `DOCTOR_API_URL` (and figure we'd have this value be the staging url for the `dev`/non-release charts...

arschles, line 32: good idea. fixed in https://github.com/deis/charts/pull/268/commits/204671dd6f49f3c42daa3c1758b83404bee94010

vdice, line 32: cool, thanks @arschles ; fyi created https://github.com/deis/deisrel/pull/63 to track the deisrel addition

arschles, line 32: thx @vdice. will mark on that PR that it requires this PR

Issue #270, opened 2016-05-25
GCS uses wrong bucket for database storage


mboersma:
There is a typo in the tpl/deis-objectstorage-secret.yaml file in workflow-dev (and workflow-rc1): the GCS database bucket has the same value as the registry bucket.

mboersma:
Closed by #271 and inline changes to the RC1 PR.

PR #271, opened 2016-05-25
fix(templates): adjust templates options accordingly

smothiki:
made the appropriate changes in the RC1 PR


smothiki:
cc @mboersma @kmala

smothiki:
closes https://github.com/deis/charts/issues/270

PR #278, opened 2016-06-01
feat(builder): add git push lock timeout env to builder rc

smothiki:
timeout int in minutes

### DIFF
--- a/workflow-dev/tpl/deis-builder-rc.yaml
+++ b/workflow-dev/tpl/deis-builder-rc.yaml
@@ -32,6 +32,9 @@ spec:
               value: "2223"
             - name: BUILDER_STORAGE
               value: "{{ or (env "STORAGE_TYPE") (.storage)}}"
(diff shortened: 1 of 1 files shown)


workflow-dev/tpl/deis-builder-rc.yaml
```diff
@@ -32,6 +32,8 @@ spec:
               value: "2223"
             - name: BUILDER_STORAGE
               value: "{{ or (env "STORAGE_TYPE") (.storage)}}"
+            - name: "GIT_LOCK_TIMEOUT"
+              value: "10"
```
sgoings, line 37: Seconds? Minutes?

smothiki, line 37: minutes

sgoings, line 37: Maybe add a comment in this area with that info?

smothiki, line 37: A lock makes sure that we are not pushing concurrently to the same GIT repo untill the previous one is finished which takes of the order minutes

PR #281, opened 2016-06-06
chore(various):	remove beta status

vdice:
Including docs update in `README.md` and version string in
the `Chart.yaml` for the `workflow-dev(-e2e)` charts.

Ref https://github.com/deis/workflow/issues/285


arschles:
Does this need to wait until 2.0 images are released?

vdice:
@arschles good question, though in this case no -- the `version` change is just a metadata change.

bacongobbler:
I think considering that we're planning on releasing 2.0 tomorrow, it's okay to remove the beta notices so it's one less step to follow for the release checklist. No opinion on this though

arschles:
@bacongobbler @vdice sounds good to me - I'm ok with merging this then 👍

PR #284, opened 2016-06-09
ref(s3): keep the access key and secret key empty by default

kmala:
It would be nice if can just leave them as empty instead of removing the existing value when user want's to use the instance profile option

### DIFF
--- a/workflow-dev/tpl/generate_params.toml
+++ b/workflow-dev/tpl/generate_params.toml
@@ -25,8 +25,10 @@ pullPolicy = "Always"
 dockerTag = "canary"
 
 [s3]
-accesskey = "YOUR KEY HERE"
-secretkey = "YOUR SECRET HERE"
+# Your AWS access key. Leave it empty if you want to use IAM credentials.
+accesskey = ""
+# Your AWS secret key. Leave it empty if you want to use IAM credentials.
+secretkey = ""
 # Any S3 region
 region = "us-west-1"
 # Your buckets.


arschles:
@kmala does this PR depend on any others?

workflow-dev/tpl/generate_params.toml
```diff
@@ -25,8 +25,10 @@ pullPolicy = "Always"
 dockerTag = "canary"
 
 [s3]
-accesskey = "YOUR KEY HERE"
-secretkey = "YOUR SECRET HERE"
+#Your aws access key.Leave it empty if you are on aws and want to use instance profile
```
bacongobbler: perhaps this would be better re-worded as

> ... if you are on AWS and want to use IAM credentials

Since more people understand what IAM is rather than "instance profile". also AWS should be capitalized

kmala:
@arschles no this doesn't depend on any other

workflow-dev/tpl/generate_params.toml
bacongobbler: also needs a space between the period and the "Leave", as well as one after the comment hash mark for punctuation.

kmala: Updated.

workflow-dev/tpl/generate_params.toml
```diff
@@ -25,8 +25,10 @@ pullPolicy = "Always"
 dockerTag = "canary"
 
 [s3]
-accesskey = "YOUR KEY HERE"
-secretkey = "YOUR SECRET HERE"
+#Your aws access key. Leave it empty if you are on AWS and want to use IAM credentials
```
mboersma: This could still be cleaned up a bit, with a leading space and a final period:
`# Your AWS access key. Leave it empty if you want to use IAM credentials.`
Same for below.

But not a showstopper, just a suggestion. 😄

kmala: Updated.

PR #285, opened 2016-06-09
chore(workflow/router-dev/Chart.yaml): prepend description with chart name

vdice:
Both to give a bit more helpful context and to enable automated updating of a chart
for a major release (where we want to drop the testing string.)

### DIFF
--- a/router-dev/Chart.yaml
+++ b/router-dev/Chart.yaml
@@ -1,7 +1,7 @@
 name: router-dev
 home: https://github.com/deis/router
 version: v2.0.0
-description: For testing only!
+description: Deis Router (For testing only!)
 maintainers:
 - Deis Team <engineering@deis.com>
 details: |-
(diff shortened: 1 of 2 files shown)


router-dev/Chart.yaml
```diff
@@ -10,8 +10,8 @@ details: |-
   The Deis Router is a simple program that manages Nginx and Nginx configuration for managing
   ingress to web applications in a Kubernetes cluster without requiring each to have their own
   external (to the cluster) load balancer.  Router(s) will require a single external load balancer,
-  handle all ingress, and direct all incoming traffic to appropriate services within the cluster.  
-  
+  handle all ingress, and direct all incoming traffic to appropriate services within the cluster.
```
arschles, line 13: maybe replace "services" with "applications" to be more specific about what the router is directing traffic to?

I know this is outside the scope of this PR, so certainly not blocking

vdice, line 13: I'll leave this for a different PR, if that's okay.  fwiw, this was an auto-whitespace change from my atom editor that I can remove if preferred...

PR #286, opened 2016-06-10
feat(database): Add easier off-cluster db config

krancour:
I've always liked running "stateless" clusters-- i.e. not only having object storage off cluster in the likes of s3 _but_ also keeping the database off-cluster in the likes of RDS, where I feel more confident I'm never going to lose data. If history is any indicator, many operators feel this way.

Although this has been _possible_ with v2 / Workflow, it has not necessarily been _easy_ to configure. (In comparison, for instance, to the ease with which off-cluster object storage has been configurable.)

This PR permits off-cluster db connection details to be specified in `generate_params.toml` and utilizes the same pattern as was used for minio in order to exclude the database RC, service, and service account from creation if they are not needed.


mboersma:
screen shot 2016-06-10 at 9 27 30 pm

Testing please ignore.

mboersma:
screen shot 2016-06-10 at 9 35 54 pm

More testing, please ignore this. Sorry!

workflow-dev/tpl/storage.sh
```diff
@@ -0,0 +1,21 @@
...
+fi
+
+db_location=`grep "db_location\s*=\s*" $HELM_GENERATE_DIR/tpl/generate_params.toml | cut -d "=" -f2 | tr -dc '[:alnum:]-\n\r'`
+if [ "$DB_LOCATION" = "off-cluster" ] || [ "$db_location" = "off-cluster" ]; then
+  rm -rf $HELM_GENERATE_DIR/manifests/deis-database-*
+else
+  helmc tpl -d $HELM_GENERATE_DIR/tpl/generate_params.toml -o $HELM_GENERATE_DIR/manifests/deis-database-rc.yaml $HELM_GENERATE_DIR/tpl/deis-database-rc.yaml
+  cp $HELM_GENERATE_DIR/tpl/deis-database-service* $HELM_GENERATE_DIR/manifests/
+fi
+helmc tpl -d $HELM_GENERATE_DIR/tpl/generate_params.toml -o $HELM_GENERATE_DIR/manifests/deis-database-secret-creds.yaml $HELM_GENERATE_DIR/tpl/deis-database-secret-creds.yaml
```
arschles, line 21: is there any reason not so split these scripts up?

arschles:
@krancour since it's fine if the scripts in https://github.com/deis/charts/pull/286#discussion-diff-67062208 are together, I'm LGTMing this

bacongobbler:
still have to manually test, but code LGTM. Philosophically I'm wondering if it might make sense to create a `workflow-dev-stateless` chart so we don't have to do all this logic for minio/postgres, though.

krancour:
@arschles the scripts are together because of a helmc limitation that (I think) exists. The commented metadata on the  first line of the toml file can only kick off ONE generator, so far as I know.

krancour:
@bacongobbler, a separate chart for stateless sounds appealing to me, but then after thinking more...

Having it separate from the usual chart doesn't reduce the work that a user of the usual chart needs to do (i.e. they don't need to fill in any optional config anyway) and it doesn't reduce the work that a user of the stateless chart needs to do (i.e.they still _do_ need to complete optional config). Since it doesn't make life easier for anyone, we should probably leave it alone.

jchauncey:
My thoughts are if we can keep just 1 chart and do stateless that would be best. Makes the release easier and makes stateless a first class citizen

jchauncey:
I wonder how we could have an e2e stateless setup?

krancour:
Just in case it was lost in my wordiness... ultimately, I'm _not_ in favor of a separate chart for this.

Re: running e2e against this, I have lots of thoughts on that topic... more than I can type on my phone. Let's discuss tomorrow.

krancour:
Removing awaiting review label, as I am aware of a small tweak this needs.

krancour:
I've amended the PR so that, as in the case of off-cluster object storage, the connection to an off-cluster DB can optionally be defined using environment variables only, i.e. without touching `genrate_params.toml`. Personally, I prefer configuring `generate_params.toml` (which is still possible, of course). This amendment is purely for the sake of consistency with the two options available for configuring off-cluster object storage.

krancour:
Complementary docs: https://github.com/deis/workflow/pull/345

vdice:
This was manually tested with an RDS instance both via the environment variable approach and via the `generate_params.toml` approach, verifying sample apps created in the first install were present on the second install (same db, naturally). Docs followed were in accompanying PR deis/workflow#345

PR #287, opened 2016-06-13
feat(workflow-dev): Add nsq to the workflow chart

jchauncey:
NSQ is a distributed messaging system build by bit.ly
It will provide us with the foundation to perform asynchronous communication across the platform. Especially in areas like Logging and Monitoring.

### DIFF
--- a/workflow-dev/tpl/deis-nsqd-rc.yaml
+++ b/workflow-dev/tpl/deis-nsqd-rc.yaml
@@ -1,0 +1,48 @@
+#helm:generate helm tpl -d $HELM_GENERATE_DIR/tpl/generate_params.toml -o $HELM_GENERATE_DIR/manifests/deis-nsqd-rc.yaml $HELM_GENERATE_FILE
+apiVersion: v1
+kind: ReplicationController
+metadata:
+  name: deis-nsqd
+  namespace: deis
+  labels:
+    heritage: deis
+spec:
+  replicas: 1
+  selector:
+    app: deis-nsqd
+  template:
+    metadata:
+      labels:
+        app: deis-nsqd
+    spec:
+      serviceAccount: deis-nsqd
+      containers:
+        - name: deis-nsqd
+          image: quay.io/{{.nsqd.org}}/nsqd:{{env "NSQ_GIT_TAG" | default .nsqd.dockerTag}}
+          imagePullPolicy: {{.nsqd.pullPolicy}}
+          command:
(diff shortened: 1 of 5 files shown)


workflow-dev/tpl/deis-nsqd-rc.yaml
```diff
@@ -0,0 +1,49 @@
...
+      labels:
+        app: deis-nsqd
+    spec:
+      serviceAccount: deis-nsqd
+      containers:
+        - name: deis-nsqd
+          image: quay.io/{{.nsqd.org}}/nsqd:{{env "NSQ_GIT_TAG" | default .nsqd.dockerTag}}
+          imagePullPolicy: {{.nsqd.pullPolicy}}
+          command:
+          - /opt/nsq/bin/start-nsqd
```
arschles, line 24: feels like this would be more flexible if the start script were on the `PATH`

arschles:
I think my comment at https://github.com/deis/charts/pull/287#discussion_r66883856 is not blocking for this PR. Code LGTM

bacongobbler:
so the upgrade path moving forward will be ok with this change? all existing data will still exist, it's just the communication between logger and fluentd right? :)

jchauncey:
Re upgrade path: that data is already lost when we upgrade so nothing to
consider on that for now. However we should look into making it more robust
in the future.

workflow-dev/tpl/deis-nsqd-rc.yaml
jchauncey, line 24: so i toyed with putting it in `/bin` but felt this was more in line with how we do our other images.

krancour, line 24: I agree that this is consistent with how many of the other images look.

workflow-dev/tpl/deis-nsqd-rc.yaml
```diff
@@ -0,0 +1,49 @@
+#helm:generate helm tpl -d $HELM_GENERATE_DIR/tpl/generate_params.toml -o $HELM_GENERATE_DIR/manifests/deis-nsqd-rc.yaml $HELM_GENERATE_FILE
+apiVersion: v1
+kind: ReplicationController
+metadata:
+  name: deis-nsqd
+  namespace: deis
+  labels:
+    heritage: deis
+    heritage: deis
```
krancour: Double `heritage` label?

PR #288, opened 2016-06-14
chore(logger/fluentd): Update logging stack to reflect nsq usage

jchauncey:
depends on the following prs:

https://github.com/deis/fluentd/pull/28
https://github.com/deis/logger/pull/85
https://github.com/deis/monitor/pull/110

The monitor PR is in progress as I am waiting for the telegraf guys to let me know when my PR will be merged. We may have to stop gap build telegraf from source and pull the binary down from binary storage if it is going to be a while.


workflow-dev/manifests/deis-logger-svc.yaml
```diff
@@ -11,10 +11,5 @@ spec:
   - port: 80
     name: http
     targetPort: http
-  - port: 514
-    name: transport
-    targetPort: 1514
-    protocol: UDP
   selector:
     app: deis-logger
-
```
krancour, line 20: Should put the empty line back at the end of the file. Almost universally, we include a blank line at the end of files. Don't ask me why, but this is definitely an outlier.

krancour:
Apart from the nit that is https://github.com/deis/charts/pull/288#discussion_r67203806, this LGTM.

workflow-dev/manifests/deis-logger-svc.yaml
jchauncey, line 20: this should be fixed now

PR #290, opened 2016-06-14
feat(swift): add support for swift storage

kmala:
ref https://github.com/deis/workflow/issues/306

### DIFF
--- a/workflow-dev/tpl/deis-objectstorage-secret.yaml
+++ b/workflow-dev/tpl/deis-objectstorage-secret.yaml
@@ -24,6 +24,14 @@ data: {{ if or (eq (env "STORAGE_TYPE") "gcs") (eq .storage "gcs")}}
   region: {{ or (env "S3_REGION") (.s3.region) | b64enc }}
   registry-bucket: {{ or (env "AWS_REGISTRY_BUCKET") (.s3.registry_bucket) | b64enc }}
   database-bucket: {{ or (env "AWS_DATABASE_BUCKET") (.s3.database_bucket) | b64enc }}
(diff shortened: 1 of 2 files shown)


workflow-dev/tpl/deis-objectstorage-secret.yaml
```diff
@@ -24,6 +24,14 @@ data: {{ if or (eq (env "STORAGE_TYPE") "gcs") (eq .storage "gcs")}}
   region: {{ or (env "S3_REGION") (.s3.region) | b64enc }}
   registry-bucket: {{ or (env "AWS_REGISTRY_BUCKET") (.s3.registry_bucket) | b64enc }}
   database-bucket: {{ or (env "AWS_DATABASE_BUCKET") (.s3.database_bucket) | b64enc }}
-  builder-bucket: {{ or (env "AWS_BUILDER_BUCKET") (.s3.builder_bucket) | b64enc }}{{else if or (eq (env "STORAGE_TYPE") "minio") (eq .storage "minio")}}
+  builder-bucket: {{ or (env "AWS_BUILDER_BUCKET") (.s3.builder_bucket) | b64enc }}{{ else if or (eq (env "STORAGE_TYPE") "swift") (eq .storage "swift")}}
+  username: {{ or (env "SWIFT_USERNAME") (.swift.username) | b64enc }}
```
bacongobbler, line 28: should these be `swift-username` so it's not a catch-all key?

kmala, line 28: no..i want it to be just username as we aren't adding provider names for other also and it would be easier/simpler to use the same name which want as environment variable if possible as it simplifies in the code of different components.

PR #293, opened 2016-06-21
feat(logger): Add redis instance for use by logger

krancour:
Depends on https://github.com/deis/logger/pull/88


krancour:
Only using an _off-cluster_ redis instance (e.g. elasticache) requires manual testing.

workflow-dev/tpl/deis-logger-rc.yaml
```diff
@@ -19,6 +19,22 @@ spec:
...
+        env:
+        - name: STORAGE_ADAPTER
+          value: redis
+{{- if or (eq (env "LOGGER_REDIS_LOCATION") "off-cluster") (eq .logger_redis_location "off-cluster") }}
+        - name: DEIS_LOGGER_REDIS_DB
+          value: "{{ or (env "LOGGER_REDIS_DB") .loggerRedis.db }}"
+        - name: DEIS_LOGGER_REDIS_SERVICE_HOST
+          value: "{{ or (env "LOGGER_REDIS_HOST") .loggerRedis.host }}"
+        - name: DEIS_LOGGER_REDIS_SERVICE_PORT
+          value: "{{ or (env "LOGGER_REDIS_PORT") .loggerRedis.port }}"
```
arschles: is this different from `{{env "LOGGER_REDIS_PORT"  | default .loggerRedis.port}}`?

arschles:
I think this looks good, so LGTMing it. I did not test off-cluster redis, however

workflow-dev/tpl/deis-logger-rc.yaml
krancour: I assume you meant to put double pipe chars there, but if so... I will try. If you're right, I think it looks cleaner.

krancour: Oh... I see you actually _did_ mean a single pipe here. I'd read it wrong. Sorry. This seems to work nicely. I will update accordingly.

krancour:
Amended with some cleaner syntax in templates and some bug fixes.

jchauncey:
When we merge this PR we should hook up redis to telegraf and create dashboards in grafana

arschles:
The following should occur before we merge this PR:
- The monitoring stack should collect redis metrics (Telegraf has a redis input plugin cc/ @jchauncey)
- We should add documentation indicating that logger instances (configured to write to redis) can be added w/o messing up log persistence 

... as per offline discussion

jchauncey:
Few things:
 1) Add redis to the list of installed components in the Chart.yaml
 2) Document the environment variables that need to be set for off-cluster storage

jchauncey:
@krancour this is ready to go

sgoings:
@krancour I made a comment on https://github.com/deis/monitor/pull/119#discussion_r70464486... I'd like to hear your thoughts.

PR #297, opened 2016-06-24
ref(workflow-dev): remove port 9090 from the service

bacongobbler:
port 9090 is only used internally for healthchecks, and as far as I know no other components
have a reliance on router's uptime for their healthchecks. Therefore we should keep this internal and not expose it to the ELB.

ping @krancour for correctness. Are you aware of any other platform components that rely on the router's port 9090 being exposed in the service?

### DIFF
--- a/workflow-dev/manifests/deis-router-service.yaml
+++ b/workflow-dev/manifests/deis-router-service.yaml
@@ -21,6 +21,3 @@ spec:
     - name: builder
       port: 2222
       targetPort: 2222
-    - name: healthz
-      port: 9090
-      targetPort: 9090


bacongobbler:
For context, this was spawned from the conversation in #community this morning: https://deis.slack.com/archives/community/p1466790066001057

krancour:
@bacongobbler this is a tricky one.

I think we _do_ want the pod's port 9090 mapped to a node port because some people might like to edit their external load balancer's healthchecks to use `HTTP GET /healthz` (instead of TCP) and that should be done against that node port (which will offer a guarantee that PROXY protocol is never used, since LB healthchecks may not support that; ELBs certainly do not).

On the flip side, there's no reason for the external load balancer (e.g. ELB) to expose this port to the rest of the world.

So we're caught up on a limitation of k8s here-- trapped between two inseparable behaviors where we want one and do not want the other.

Personally, I would opt for closing this.

bacongobbler:
Given that port 9090 is nothing but a router healtcheck, I agree.

PR #298, opened 2016-06-27
fix(minio):create minio when using storage type as minio via env vars

kmala:


workflow-dev/tpl/storage.sh
```diff
@@ -1,9 +1,7 @@
...
 store=`grep "storage\s*=\s*" $HELM_GENERATE_DIR/tpl/generate_params.toml | cut -d "=" -f2 | tr -dc '[:alnum:]\n\r'`
-if [ -n "$STORAGE_TYPE" ]; then
-  if [ "$STORAGE_TYPE" != "minio" ]; then
-    rm -rf $HELM_GENERATE_DIR/manifests/deis-minio-*
-  fi
+if [ -n "$STORAGE_TYPE" ] && [ "$STORAGE_TYPE" != "minio" ]; then
+  rm -rf $HELM_GENERATE_DIR/manifests/deis-minio-*
 elif [ "$store" != "minio" ]; then
   rm -rf $HELM_GENERATE_DIR/manifests/deis-minio-*
 else
```
krancour: Note on positioning of this comment... I cannot place it any further down (past the diff), but I mean for this to apply to the _entire_ `if / elif / else / fi` block.

This logic can all be expressed more succinctly and with less confusion. Currently, there are two paths to the same result. (That result being `rm -rf $HELM_GENERATE_DIR/manifests/deis-minio-*`.) Also, currently, there's a lot of "negative logic." So for a human reading the code, it's backwards from how people usually think. It's starting with "if something isn't true; otherwise..." instead of starting out with "if something _is_ true; otherwise..."

I would suggest the following:

```
if [ "$STORAGE_TYPE" = "minio" ] || [ "$store" = "minio" ]; then
  helmc tpl -d $HELM_GENERATE_DIR/tpl/generate_params.toml -o $HELM_GENERATE_DIR/manifests/deis-minio-rc.yaml $HELM_GENERATE_DIR/tpl/deis-minio-rc.yaml
  cp $HELM_GENERATE_DIR/tpl/deis-minio-service* $HELM_GENERATE_DIR/manifests/
else
  rm -rf $HELM_GENERATE_DIR/manifests/deis-minio-*
fi
```

kmala: I understand what you are trying to say but we can't move away from negative logic as the default value present in the toml file is minio.

krancour: I see what you're saying... as long as the toml has not been edited, that first line is _always_ going to resolve as true, no matter what the env var is set to.

This will work:

```
if [ "$STORAGE_TYPE" = "minio" ] || ( [ -z "$STORAGE_TYPE" ] && [ "$store" = "minio" ] ); then
  helmc tpl -d $HELM_GENERATE_DIR/tpl/generate_params.toml -o $HELM_GENERATE_DIR/manifests/deis-minio-rc.yaml $HELM_GENERATE_DIR/tpl/deis-minio-rc.yaml
  cp $HELM_GENERATE_DIR/tpl/deis-minio-service* $HELM_GENERATE_DIR/manifests/
else
  rm -rf $HELM_GENERATE_DIR/manifests/deis-minio-*
fi
```

That will at least eliminate having two conditions leading to the same `rm -rf $HELM_GENERATE_DIR/manifests/deis-minio-*`

kmala: @krancour updated

PR #302, opened 2016-07-06
wip: mirror workflow release artifacts to private registry

slack:
Two parter. Spoke with some folks at Dockercon that needed to run Workflow in a internet-restricted enviroment. External access _must_ go through a proxy. They do have access to a private registry, so I wanted to see what it would take to make it much easier for someone to deploy Workflow in that environment.

First, create a little script to pull, re-tag and push release images to a different registry endpoint.

Second, re-work the charts to reference a defaults section for registry, org and pullPolicy values.

As I haven't gone through the release dance, I wanted thoughts on the general approach before completing the individual charts.

Does the chart refactorsimplify the relase tooling? Does it complicate workflow-dev and git-sha images?

### DIFF
--- a/_scripts/mirror-images.sh
+++ b/_scripts/mirror-images.sh
@@ -1,0 +1,46 @@
+#!/bin/bash
+
+# this script is built to pull Deis Worfklow release images, re-tag them and
+# subsequently push the images to a new repository
+
+# Usage: MIRROR_DESTINATION=quay.io/jhansen ./_script/mirror-images.sh
+
+# Override the version and source locations by setting the following
+# environment variables:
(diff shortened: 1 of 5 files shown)


mboersma:
@slack is this still in-progress, or should we review it now?

slack:
I'm not actively working on this, but I'm still curious about what folks think of the `generate_params.toml` changes.

If folks like the params refactor, I'm happy to finish it up! Just not sure about the implications for ci/automation/deisrel.

As for the image mirroring script, I haven't heard back from the folks I spoke to at Dockercon, so we could simply drop that bit.

slack:
Closing! We can revisit it later if we wish.

PR #309, opened 2016-07-19
feat(test): add TEST env variable to workflow e2e chart

smothiki:
make Ginkgo tests configurable via TEST env var, as workflow-e2e tests has more test suites piling up in the future this comes in handy.
refer https://github.com/deis/workflow-e2e/pull/263

### DIFF
--- a/workflow-dev-e2e/tpl/workflow-e2e-pod.yaml
+++ b/workflow-dev-e2e/tpl/workflow-e2e-pod.yaml
@@ -19,6 +19,9 @@ spec:
         value: "true"
       - name: CLI_VERSION
         value: "{{env "CLI_VERSION" | default "latest"}}"
+       # set TEST env variable to run appropriate tests in e2e suite
+      - name: TEST
+        value: "{{env "TEST" | default "e2e"}}"
     volumeMounts:
     - name: artifact-volume
       mountPath: /root


workflow-rc2-e2e/tpl/workflow-e2e-pod.yaml
```diff
@@ -17,6 +17,8 @@ spec:
         value: "{{env "GINKGO_NODES" | default 15}}"
       - name: JUNIT
         value: "true"
+      - name: TEST
+        value: "{{env "TEST" | default e2e}}"
```
vdice: This could technically just be `value: "{{env "TEST"}}"` per https://github.com/smothiki/workflow-e2e/blob/9a55c84acdc224590b225195bb00b6552ac71981/docker-test-integration.sh#L17-L22 right?  (In that link, it looks like the default case is anything other than `TEST='bps'` at this point)

workflow-v2.1.0-e2e/tpl/workflow-e2e-pod.yaml
```diff
@@ -17,6 +17,8 @@ spec:
         value: "{{env "GINKGO_NODES" | default 15}}"
       - name: JUNIT
         value: "true"
+      - name: TEST
```
sgoings: Could this get a comment or explanation as to its purpose?

smothiki: I'll change the description.

smothiki: Updated the description to reflect this.

smothiki:
tested yaml syntax manually 

```
$ rm -rf ~/.helmc/workspace/charts/workflow-dev-e2e
smothikis-MacBook-Pro:charts smothiki$ cp -r workflow-dev-e2e ~/.helmc/workspace/charts/.
smothikis-MacBook-Pro:charts smothiki$ helmc generate workflow-dev-e2e
---> Ran 1 generators.
```

PR #313, opened 2016-07-21
chore(*): remove old charts that reference pre 2.0.0 release

sgoings:
I had to make up for my lack of coding at some point...


bacongobbler:
I think it's reasonable at this point to remove the old charts. LGTM!

mboersma:
> I think it's reasonable at this point to remove the old charts.

And if someone really needs to resurrect the deis-logger-beta1 chart, for example, it's still here in the `git` history, and the Docker images will remain at quay.io for the foreseeable future.

PR #314, opened 2016-07-22
feat(registry):Add support for private registry

kmala:


workflow-dev/tpl/deis-registry-secret.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: registry-secret
+  namespace: deis
+  labels:
+    heritage: deis
+  annotations:
+    deis.io/objectstorage: "{{ or (env "REGISTRY_LOCATION") (.registry_location)}}"
```
slack: Should this annotation be `deis.io/registry-location` or something?

workflow-dev/tpl/deis-registry-secret.yaml
```diff
@@ -0,0 +1,22 @@
...
+type: Opaque
+data: {{ if or (eq (env "REGISTRY_LOCATION") "gcs") (eq .registry_location "gcr")}}
+  key.json: {{ or (env "GCR_KEY_JSON") (.gcr.key_json) | b64enc}}
+  server: {{ or (env "GCR_SERVER") (.gcr.server) | b64enc}}{{ else if or (eq (env "REGISTRY_LOCATION") "ecr") (eq .registry_location "ecr")}}
+  accesskey: {{ or (env "ECR_ACCESS_KEY") (.ecr.accesskey) | b64enc }}
+  secretkey: {{ or (env "ECR_SECRET_KEY") (.ecr.secretkey) | b64enc }}
+  region: {{ or (env "ECR_REGION") (.ecr.region) | b64enc }}
+  registryid: {{ or (env "ECR_REGISTRY_ID") (.ecr.registryid) | b64enc }}
+  server: {{ or (env "ECR_SERVER") (.ecr.server) | b64enc }}{{else if or (eq (env "REGISTRY_LOCATION") "off-cluster") (eq .registry_location "off-cluster")}}
```
slack: I think we should use `hostname` here, even though the docker login command uses `--server`

workflow-dev/tpl/deis-registry-secret.yaml
```diff
@@ -0,0 +1,22 @@
...
+  server: {{ or (env "REGISTRY_SERVER") (.off_cluster_registry.server) | b64enc }}
+  namespace: {{ or (env "REGISTRY_NAMESPACE") (.off_cluster_registry.namespace) | b64enc }}
```
slack: Namespace feels confusing since it overlaps with Kubernetes namespaces... Would something like `registry_org`... help?

workflow-dev/tpl/deis-registry-token-refresher.yaml
```diff
@@ -0,0 +1,37 @@
...
+    app: deis-registry-token-refresher
+  template:
+    metadata:
+      labels:
+        heritage: deis
+        app: deis-registry-token-refresher
+    spec:
+      containers:
+        - name: deis-registry-token-refresher
+          image: kmala/registry-token-refresher:canary
```
helgi: Is there a deis image?

kmala: currently no...i will change it once this PR https://github.com/deis/registry-token-refresher/pull/1 is merged

workflow-dev/tpl/deis-registry-rc.yaml
```diff
@@ -1,4 +1,3 @@
-#helm:generate helm template -o $HELM_GENERATE_DIR/manifests/deis-registry-rc.yaml -d $HELM_GENERATE_DIR/tpl/generate_params.toml $HELM_GENERATE_DIR/tpl/deis-registry-rc.yaml
```
mboersma, line 1: If this file isn't a template, should it be moved under the manifests/ dir?

_Update_: never mind, I didn't see the entire context in the diff view.

workflow-dev/tpl/deis-registry-secret.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: registry-secret
+  namespace: deis
+  labels:
+    heritage: deis
+  annotations:
+    deis.io/registry-location: "{{ or (env "REGISTRY_LOCATION") (.registry_location)}}"
```
mboersma: Nitpick: could you put a space before the closing double-braces `}}`? I think go templating reads more easily that way but we're not being consistent here, or on some of the following lines.

workflow-dev/tpl/deis-registry-rc.yaml
kmala, line 1: this is still a template just that it should be generated only when the registry_location is on-cluster.The generate logic is present in the storage.sh script.

workflow-dev/tpl/deis-registry-token-refresher.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: v1
+kind: ReplicationController
```
helgi: Should this be a Deployment instead?

kmala: yes...the charts needs refactor.

PR #315, opened 2016-07-26
Move a few components to use Deployments

helgi:
RC for deis-router is kept around so it can be taken out of rotation without needing manual intervention (see replicas 0 and changed labels)


helgi:
Moved a few more components

helgi:
Closing in favour of https://github.com/deis/charts/pull/316

PR #316, opened 2016-07-26
feat(*): Move a few components to Deployments

helgi:
RC for affected components are kept around so it can be taken out of rotation without needing manual intervention (see replicas 0 and changed labels)


helgi:
Does anyone have thoughts on more components to move to Deployments ?

kmala:
What is the reason behind database not including in this move?

helgi:
Deployments are not meant for persistent data, PetSets are what we want for that.

Using Deployments you will have overlapping masters when dealing with something like a Database

workflow-dev/tpl/deis-registry-deployment.yaml
```diff
@@ -0,0 +1,60 @@
...
+spec:
+  replicas: 1
+  strategy:
+    rollingUpdate:
+      maxSurge: 1
+      maxUnavailable: 0
+    type: RollingUpdate
+  selector:
+    matchLabels:
+    app: deis-registry
```
kmala: not properly indented.

kmala:
i had tested manually and its working properly.

PR #325, opened 2016-08-04
ref(registry_location): Add support to set registry location through env

kmala:


workflow-dev/tpl/deis-builder-deployment.yaml
```diff
@@ -46,7 +46,7 @@ spec:
             - name: BUILDER_STORAGE
               value: "{{ or (env "STORAGE_TYPE") (.storage)}}"
             - name: "DEIS_REGISTRY_LOCATION"
-              value: "{{ .registry_location }}"
+              value: "{{ or (env "REGISTRY_LOCATION") (.registry_location) }}"
```
arschles, line 49: why not the familiar pattern of `{{ (env "REGISTRY_LOCATION") | default .registry_location }}`?

same with the below

arschles, line 49: since this appears to work anyway, I'll consider it non-blocking for the PR

PR #332, opened 2016-08-16
feat(minio): Add option to use hostPath for minio storage

tmc:
This adds an option to minio to set up a hostPath volume. Running deis in minikube is a use case.

### DIFF
--- a/workflow-dev/tpl/deis-minio-rc.yaml
+++ b/workflow-dev/tpl/deis-minio-rc.yaml
@@ -44,10 +44,19 @@ spec:
           args:
             - "server /home/minio/"
           volumeMounts:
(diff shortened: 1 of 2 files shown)


helgi:
You'll want to do those changes against `workflow-dev`. That's the chart we use as the basis for releases

bacongobbler:
This will only work with single-node clusters where there is only one worker node available. As long as we're abundantly clear about that in the docs I'm OK with this.

rimusz:
the option of `nodeSelector` should be there as well if the k8s cluster is not the single-node one.

bacongobbler:
we've stopped development on helm-classic. Please re-open another PR on deis/minio against that chart. Thanks!

PR #341, opened 2016-08-19
ref(workflow-dev): change container port for the router to 80 and 443

bacongobbler:
see deis/router#239


krancour:
Why is this changing? Did the ports within the router container change? If they did, I have to assume the router is running as root now, and I would have to once again articulate the security concerns that arise from running containerized processes as root.

bacongobbler:
Yes, and I explained my (wip) thoughts in deis/router#239.

PR #342, opened 2016-08-20
chore(workflow-v2.4.1): releasing workflow-v2.4.1

helgi:
Only includes https://github.com/deis/controller/releases/tag/v2.4.1

This was all based off v2.4.0 instead of workflow-dev


workflow-v2.4.1-e2e/Chart.yaml
```diff
@@ -0,0 +1,13 @@
+name: workflow-v2.4.0-e2e
```
mboersma: Should be changed to `2.4.1` here, and below.

workflow-v2.4.1-e2e/README.md
```diff
@@ -0,0 +1,12 @@
+# Workflow 2.4.0 e2e Tests (parallel)
+
+End-to-end tests for the Deis Workflow open source PaaS, executed in parallel.
+
+NOTE: This assumes a fresh `helmc install` of Deis v2. The test suite creates
+an initial user with admin privileges (who is deleted when tests complete).
+See https://github.com/deis/workflow-e2e for more detail.
+
+```console
+$ helmc uninstall -n deis -y workflow-v2.4.0-e2e ; helmc install workflow-v2.4.0-e2e
```
mboersma: 2.4._1_

workflow-v2.4.1-e2e/tpl/generate_params.toml
```diff
@@ -0,0 +1,5 @@
+[e2e]
+org = "deis"
+dockerTag = "v2.4.0"
```
mboersma, line 3: Is there a new e2e docker image tag? Probably ok to leave this as 2.4.1 if not.

helgi, line 3: You mean leave as 2.4.0? I didn't make a new one since originally we talked about not doing e2e but turns out it was a requirement of sorts

PR #345, opened 2016-08-29
fix(imagepullpolicy): pass the image pull policy of slugbuilder,dockerbuilder and slugrunner

kmala:


workflow-dev/tpl/deis-builder-deployment.yaml
```diff
@@ -65,8 +65,12 @@ spec:
               value: "10"
             - name: "SLUGBUILDER_IMAGE_NAME"
               value: "quay.io/{{.slugbuilder.org}}/slugbuilder:{{ env "SLUGBUILDER_GIT_TAG" | default .slugbuilder.dockerTag }}"
+            - name: SLUG_BUILDER_IMAGE_PULL_POLICY
+              value: {{ .slugbuilder.pullPolicy }}
```
sgoings, line 69: Where are the defaults for these pullPolicies?

kmala: `Always` for dev and `IfNotPresent` for release charts

Issue #339, opened 2016-08-17
Convert Logging stack to use deployments


jchauncey:
The logging stack should move to using deployments like the other components

Issue #338, opened 2016-08-17
Convert Monitor stack to use deployments


jchauncey:
The monitor stack should follow the move of going to the deployment resource like the other components.

PR #346, opened 2016-09-06
feat(deployments): move components to use deployments instead of RC

### LINKED ISSUE #337: We're only 1/2 way transitioned to deployments?
Looks like we're using deployments for some components (e.g. controller and router), but are still relying on RCs for others (e.g. database, logger... others?) We could stand to be more consistent on this.

(I know RCs are supposed to continue to exist and just be set to `replicas: 0`, and that's _not_ what I am reporting here. I'm RCs still set to `replicas: 1`, with no corresponding deployment existing at all.)

kmala:
fixes #337 #338 #339

### DIFF
--- a/workflow-dev/tpl/deis-database-deployment.yaml
+++ b/workflow-dev/tpl/deis-database-deployment.yaml
@@ -1,16 +1,20 @@
-apiVersion: v1
-kind: ReplicationController
+apiVersion: extensions/v1beta1
+kind: Deployment
 metadata:
   name: deis-database
   namespace: deis
   labels:
     heritage: deis
   annotations:
+    helm-keep: "true"
     component.deis.io/version: {{env "DATABASE_GIT_TAG" | default .database.dockerTag}}
 spec:
-  replicas: 0
+  replicas: 1
+  strategy:
+    type: Recreate
   selector:
-    app: deis-database
+    matchLabels:
+      app: deis-database
   template:
     metadata:
       labels:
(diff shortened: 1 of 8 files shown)


workflow-dev/tpl/deis-database-deployment.yaml
```diff
@@ -0,0 +1,60 @@
...
+  namespace: deis
+  labels:
+    heritage: deis
+  annotations:
+    helm-keep: "true"
+    component.deis.io/version: {{env "DATABASE_GIT_TAG" | default .database.dockerTag}}
+spec:
+  replicas: 1
+  strategy:
+    type: Recreate
```
helgi, line 14: For others to understand, `Recreate` is the important difference here compared to Controller / Router which use `RollingUpdate`

workflow-dev/tpl/deis-logger-deployment.yaml
```diff
@@ -0,0 +1,68 @@
...
+  namespace: deis
+  labels:
+    heritage: deis
+  annotations:
+    helm-keep: "true"
+    component.deis.io/version: {{env "LOGGER_GIT_TAG" | default .logger.dockerTag}}
+spec:
+  replicas: 1
+  strategy:
+    type: Recreate
```
helgi: Any reason the logger can't be a `RollingUpdate`?

kmala: i think if there will be an issue if two logger instances write to the same redis instance simultaneously.
cc @krancour

kmala:
@krancour can you check the logger components?
@jchauncey can you check the monitor components?
@bacongobbler can you check the database component?

bacongobbler:
database component looks good to me. My only concern before was the rolling deployment model, but recreate should work.

workflow-dev/tpl/deis-logger-deployment.yaml
jchauncey: i think they have locks in place so we can horizontally scale logger. it shouldnt be an issue

helgi: Then using the `RollingUpdate` model would be much better

kmala: if you are sure about it then i will change it to rollingupdate

PR #348, opened 2016-09-07
Release v2.5.0

mboersma:


workflow-v2.5.0/tpl/generate_params.toml
```diff
@@ -0,0 +1,249 @@
...
+
+[dockerbuilder]
+org = "deis"
+pullPolicy = "IfNotPresent"
+dockerTag = "v2.3.2"
+
+[controller]
+org = "deis"
+pullPolicy = "IfNotPresent"
+dockerTag = "v2.5.0"
```
bacongobbler: can you bump this to v2.5.1 please?

workflow-v2.5.0/tpl/generate_params.toml
```diff
@@ -0,0 +1,249 @@
...
+dockerTag = "v1.1.0"
+hostPort = 5555
+limits_cpu = "100m"
+limits_memory = "50Mi"
+
+[workflowManager]
+org = "deis"
+pullPolicy = "IfNotPresent"
+dockerTag = "v2.4.1"
+versionsApiURL = "https://versions-staging.deis.com"
```
kmala: this and below should be changed to "https://versions.deis.com" and "https://doctor.deis.com"

mboersma: I reverted them to the `-staging` URLs in a temp commit here, which is intended to be discarded before we actually merge this.

vdice:
LGTM besides a friendly reminder in https://github.com/deis/charts/pull/348/files#r78082168 which looks like @mboersma has on top of mind!

PR #351, opened 2016-09-13
ref(controller): SLUG_RUNNER_IMAGE_PULL_POLICY got removed and IMAGE_PULL_POLICY is the new thing to use

helgi:
https://github.com/deis/controller/pull/1053 for rationale

### DIFF
--- a/workflow-dev/tpl/generate_params.toml
+++ b/workflow-dev/tpl/generate_params.toml
@@ -132,9 +132,19 @@ dockerTag = "canary"
 # limits_cpu = "100m"
 # limits_memory = "50Mi"
 
+# Set the default (global) way of how Application (your own) images are
+# pulled from within the Controller.
+# This can be configured per Application as well in the Controller.
+#
+# This affects pull apps and git push (slugrunner images) apps
+#
+# Values values are:
+# - Always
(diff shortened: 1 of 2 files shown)


workflow-dev/tpl/generate_params.toml
```diff
@@ -45,6 +45,17 @@ logger_redis_location = "on-cluster"
...
+# Set the default (global) way of how Application (your own) images are
+# pulled from within the Controller.
+# This can be configured per Application as well in the Controller.
+#
+# This affects pull apps and git push (slugrunner images) apps
+#
+# Values values are:
+# - Always
+# - IfNotPresent
+image_pull_policy = "Always"
```
kmala: it would be better  to have it as part of the controller as it is related to only it

helgi: Thought about that but we already have a `pullPolicy` there which is only related to the controller image itself, which seems confusing at best if we have 2 very similarly named things there

kmala: can we change to app_pull_policy then?

helgi: Check now

PR #352, opened 2016-09-19
feat(workflow-dev-e2e): add spotify/docker-gc to e2e

bacongobbler:
As seen in our GKE clusters, the docker images for the applications spawned by e2e
are never reaped. This causes thousands of images to persist, all identical to each other on a single
node and can trigger etcd issues.

See https://github.com/deis/controller/issues/1082#issuecomment-248079239 for some background context.

### DIFF
--- a/workflow-dev-e2e/manifests/spotify-gc-daemon.yaml
+++ b/workflow-dev-e2e/manifests/spotify-gc-daemon.yaml
@@ -17,7 +17,6 @@ spec:
         heritage: deis
         app: deis-spotify-gc
     spec:
-      serviceAccount: deis-spotify-gc
       containers:
       - name: deis-spotify-gc
         image: spotify/docker-gc:latest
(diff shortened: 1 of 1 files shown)


bacongobbler:
@mboersma mentioned to me during manual testing that this is a one-off container. That is, it runs to completion then dies. It looks like the container is unhealthy in a CrashLoopBackoff but in reality it's just running a few times during the e2e job and that's all we really need.

Open to other suggestions to improve upon this, but I think this is good enough to ship so we stop breaking GKE :)

mboersma (approved):
This should be sufficient to fix the immediate problem for e2e in GKE. We can revisit later to see if the behavior can be made cleaner, since the pod continuing to `CrashLoopBackOff` and restart could cause confusion (although is actually not far away from desired `cron`-like behavior.

mboersma:
I made a trivial [fork of spotify/docker-gc](https://github.com/mboersma/docker-gc/tree/as-cron-job) that keeps the container from exiting by running the script every two hours via `crond`. You can test it using `quay.io/mboersma/spotify-gc:0.1.2` if you're curious. Maybe something like that could be included with Deis Workflow itself, although it might be better to start from `deis/base`.

PR #353, opened 2016-09-23
feat(monitor): Make grafana credentials configurable

wearethefoos:
In the docs for Grafana it states:

> The default username/password of `admin/admin` can be overridden at any time by setting the following environment variables in `$CHART_HOME/workspace/workflow-$WORKFLOW_RELEASE/manifests/deis-monitor-grafana-rc.yaml`: `GRAFANA_USER` and `GRAFANA_PASSWD`.

For us it would be great if this was configurable via the template and/or environment variables.

This pull should make both options available.

I'd gladly make a pull as well for the docs, e.g:

> The default username/password of `admin/admin` can be overridden at any time by setting the following environment variables: `GRAFANA_USER` and `GRAFANA_PASSWD` or by editing the default parameters in `generate_params.toml` before running `helmc generate`.

Let me know if this is a way to go, or if I need to do anything differently!

Cheers.

### DIFF
--- a/workflow-dev/tpl/deis-monitor-grafana-rc.yaml
+++ b/workflow-dev/tpl/deis-monitor-grafana-rc.yaml
@@ -35,6 +35,10 @@ spec:
           value: http://$(DEIS_MONITOR_INFLUXAPI_SERVICE_HOST):$(DEIS_MONITOR_INFLUXAPI_SERVICE_PORT_TRANSPORT)
         - name: "BIND_PORT"
           value: "3500"
+        - name: "DEFAULT_USER"
+          value: {{env "GRAFANA_USER" | default .grafana.user}}
+        - name: "DEFAULT_USER_PASSWORD"
+          value: {{env "GRAFANA_PASSWD" | default .grafana.password}}
         ports:
         - containerPort: 3500
           name: ui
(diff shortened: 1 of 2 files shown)


bacongobbler:
Hey @foxycoder, would you mind making this change against the workflow-dev chart instead? We intentionally keep the release charts as-is and just roll new features into workflow-dev, which will then be rolled into the next release.

wearethefoos:
@bacongobbler makes total sense! Sorry will do :)

wearethefoos:
@bacongobbler I moved the changes to workflow-dev chart. Let me know if you guys need anything else.

workflow-dev/tpl/deis-monitor-grafana-rc.yaml
```diff
@@ -35,6 +35,10 @@ spec:
           value: http://$(DEIS_MONITOR_INFLUXAPI_SERVICE_HOST):$(DEIS_MONITOR_INFLUXAPI_SERVICE_PORT_TRANSPORT)
         - name: "BIND_PORT"
           value: "3500"
+        - name: "GRAFANA_USER"
```
kmala: i think you should rename it to `DEFAULT_USER`  and `DEFAULT_USER_PASSWORD` https://github.com/deis/monitor/blob/master/grafana/rootfs/usr/share/grafana/start-grafana#L7 or change the grafana code also

mboersma:
Jenkins, add to whitelist.

workflow-dev/tpl/deis-monitor-grafana-rc.yaml
wearethefoos: Nice one @kmala. Done :)

kmala (approved):

PR #365, opened 2016-10-03
fix(charts): make grafana creds configurable and remove unused minio POD_IP env

kmala:
ref https://github.com/deis/minio/pull/119


helgi:
That the only change we missed during the RC -> Deployment switch?

kmala:
yes..that and minio but since minio change is not required i removed it from the rc also so not to confuse anyone.

PR #370, opened 2016-10-10
chore(workflow-v2.7.0): releasing workflow-v2.7.0(-e2e)

mboersma:


jchauncey:
if you want to test upgrading here ya go - https://deis.com/docs/workflow/managing-workflow/upgrading-workflow/

jchauncey:
@mboersma we should bring over the changes that are in #371

mboersma:
Updated to sync up with #371.

workflow-v2.7.0/tpl/deis-database-service.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: deis-database
+  namespace: deis
+  labels:
+    heritage: deis
+  amnotations:
```
vdice: s/amnotations/annotations

jchauncey: good catch. we should maek sure that same spelling error isnt in @kmala 's pr

jchauncey: just checked and its not.

mboersma: My bad--fixed!

PR #372, opened 2016-10-12
feat(monitor): Allow for off cluster influxdb

jchauncey:
Test steps:
- clone PR
- In generate_params.toml
  - Mark the property `influxdb_location` as `off-cluster`
  - Add values under influxdb for url, user, password, and database
- run `helm generate`
- verify that configuration values were inserted into `manifests/deis-monitor-telegraf-daemon.yaml` and `manifests/deis-monitor-grafana-deployment.yaml`


kmala (approved):

mboersma:
After `helmc generate`, the URL is enclosed by escaped quotes in `manifests/deis-monitor-telegraf-daemon.yaml`:

``` yaml
          - name: "INFLUXDB_URLS"
            value: "\"deis.rocks\""
```

But they don't show up in `manifests/deis-monitor-grafana-deployment.yaml`:

``` yaml
        - name: "INFLUXDB_SERVICE_URL"
          value: "deis.rocks"
```

Assuming that's intentional, everything looks good.

jchauncey:
So the escaped quotes are there so we actually quote the string when it gets passed to telegraf.

PR #373, opened 2016-10-12
feat(controller): Add registration mode env var

jchauncey:


jchauncey:
Alright so this makes the registration mode a little more explicit.

workflow-dev/tpl/deis-controller-deployment.yaml
```diff
@@ -56,6 +56,12 @@ spec:
 {{- end}}
 {{- end}}
           env:
+            # NOTE(jchauncey): Possible values are:
+            # enabled - allows for open registration
+            # disabled - turns off open registration
+            # admin_only - allows for registration by an admin only.
+            - name: REGISTRATION_MODE
+              value: enabled
```
kmala: can we have the value taken from the params.toml because that is where people add all the configuration options and has more visibility.

jchauncey: yup good catch

PR #375, opened 2016-10-17
fix(influx):copy the api and ui services for influx properly

kmala:


jchauncey:
thanks for catching that.

PR #383, opened 2016-11-01
fix(workflow-dev/tpl/deis-database-deployment.yaml): rename env var

vdice:
from DATABASE_GIT_TAG to POSTGRES_GIT_TAG, so that when ci sets the latter in the env, it is properly used in the chart.


bacongobbler:
Excellent catch.

PR #389, opened 2016-12-08
doc(readme): increase the deprecation visibility

slack:
Had a few folks fall down the "deis/charts" hole when trying to install workflow. Thoughts on this approach?

I had also thought about removing the development, contributing, and installation sections... That may be a bit too much!


bacongobbler:
I'd remove the development, contributing, installation etc. sections. We're no longer actively maintaining it so making the README nothing but a big fat notice that it's deprecated sounds good to me. Otherwise there will be people who will likely skip the notice and carry on to the docs. That has happened before with other deprecated projects.

mboersma:
Excellent--much clearer messaging. Thanks @slack!

